PR #104, opened 2015-06-05
Doc: Edge() initializer also accepts Node objects for src and dst.

nanno-lang:
A minor addition to documentation, triggered by comments at http://stackoverflow.com/questions/6953957/edges-between-two-subgraphs-in-pydot

### DIFF
--- a/pydot.py
+++ b/pydot.py
@@ -841,8 +841,11 @@ class Edge(object,  Common ):
     
     edge(src, dst, attribute=value, ...)
     
-    src: source node's name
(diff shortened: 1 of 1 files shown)


johnyf:
Thank you for improving the documentation, merged as 588f8dd847aecb5a7a6165eaa3a714561f6a1e11.
Also, as of 4979ea96f0177336c59d098482a11bd066f45451, I changed the API of `Edge.__init__` to be less surprising to users, and more straightforward.

PR #106, opened 2015-11-10
updated to python 3.5 and fixed MRO errors

soeltjen:
Updated all code to python 3.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,9 @@
+.vs/
+__pycache__/
+build/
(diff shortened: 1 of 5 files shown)


johnyf:
Thank you. The primary fork that addressed updates to Python 3 is https://github.com/nlhepler/pydot.
The changes here were committed there earlier, and have now been applied upstream.
Also, please note that the changes here would be insufficient for python 3 compatibility, because of issues with binary and text io.

PR #112, opened 2016-01-12
Fix python3 byte string join.

vtsingaras:


johnyf:
Thanks. This change has been made obsolete in 311642aaddc5e950792fbd49d56b676c0ac66259, by calling `subprocess.Popen.communicate` in `pydot.Dot.create`.

PR #114, opened 2016-02-15
Hardcode revision number

jbchouinard:
Since the repo was moved from SVN on google code to github, `$LastChangedRevision$` doesn't get replaced with an actual revision number.


johnyf:
Thanks, fixed by the nearly equivalent 8ca3077ece8be35bc1ba528a1d52bcc0a218fc5c.

Issue #90, opened 2015-04-13
Include tests into release tarball


GoogleCodeExporter:
Hello,
the tarball does not contain unittest as available at 
http://pydot.googlecode.com/svn/trunk/test/ which is rather unfortunate since 
they are an important verification tool (in particular for Linux Distribution 
as Debian, for which I package pydot).

Please consider including those tests into the released tarballs.

Regards,
Sandro

Original issue reported on code.google.com by `sandro.tosi` on 21 Mar 2014 at 9:05

johnyf:
Addressed by ae80697e61e9e3b265175676bb78db6396542616.

PR #120, opened 2016-04-23
Include tests in release source tarball

EdwardBetts:
A test suite is a very useful thing, it would be great to include it in the release tarball that is uploaded to pypi. That way people who download and build the release can run the tests to ensure the library is working.

Linux distributions, such as Debian, base their packages of Python modules on the pypi release. This means the test suite can be run when building a Debian package, and so catch mistakes in the packaging or errors introduced into dependencies.

I've written some more on this topic on the Debian Python mailing list: https://lists.debian.org/debian-python/2016/04/msg00074.html

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+recursive-include test *


johnyf:
Unfortunately, the current test suite is:

1. disproportionately large (about 1MB, when the rest is 300 KB)
2. includes binary images whose copyright status is unknown, and thus cannot be distributed on PyPI.

Thus, `recursive-include test *` cannot be used. Instead, the test suite needs to either be refreshed, or a representative subset of `dot` files be selected for inclusion in the archive distribution on PyPI.

The last action item is delegated to #90.

Issue #132, opened 2016-09-28
Problem with quoted strings


IARI:
Consider the following graph woith a quoted label:

```
digraph G {
    a [label="Graph with quoted dash \"-\" string"];
}
```

it should look like this:

![rendered version](http://g.gravizo.com/g?
digraph G {
    a [label="Graph with quoted dash \\"-\\" string"];
}
)

However, dot_parser doesn't recognize it, producing

> pyparsing.ParseException: Expected "}" (at char 18), (line:2, col:7)

johnyf:
A patch would be welcome.

At a higher level of planning, I think that `dot_parser` should be replaced with a PLY-based parser, because:

1. `dot_parser` anyway needs to be rewritten (witness to that are the issues reported here), and
2. `pyparsing` is for top-down parsing, which is less efficient than bottom-up parsing. So, lex-yacc (`ply`) should replace `pyparsing` in `dot_parser` (also resulting in much cleaner code).

There shouldn't be much need to write a `ply` parser for DOT from scratch, because `grandalf` contains one (https://github.com/bdcht/grandalf/blob/master/grandalf/utils/dot.py). In the worst case, it would require some rewriting. However, `grandalf` has a different license (so borrowing code involves first asking for dual licensing).

Secondly, this poses the question of package architecture: where should the DOT parser be, in `pydot`, in `grandalf`, who imports whom then? Should we merge the two packages? I think that `grandalf`'s purpose is defined somewhat differently. `grandalf` is a pure Python implementation of graph layout algorithms, so a substitute for (a subset of) GraphViz.

It would be great to call `grandalf` from here, when GraphViz is missing. In other words, it would be reasonable to regard `grandalf` as a drawing backend. So, it would be `pydot` importing `grandalf`, not the other way around. In principle they could also be merged, but the usefulness of `pydot` as it is now I think is related more to its long history, so wide use, thus the API it defines. A completely restructured API (e.g., using `networkx`) would be preferable, but that would be equivalent to writing a new package, rather than changing `pydot`. So, we could identify `pydot` with (most of) its API, and let fresh architecture be developed in other packages.
For now, I think that it would be better to focus on the parser issues, and let a pure Python drawing backend as future consideration.

[...]

PR #134, opened 2016-10-04
Issue #132: escaping quoted strings

IARI:
I hope this fixxes #132 for now.

### DIFF
--- a/dot_parser.py
+++ b/dot_parser.py
@@ -434,7 +434,7 @@ def graph_definition():
         identifier = Word(alphanums + "_." ).setName("identifier")
 
         double_quoted_string = QuotedString(
-            '"', multiline=True, unquoteResults=False)  # dblQuotedString
+            '"', multiline=True, unquoteResults=False, escChar='\\')  # dblQuotedString
 
         noncomma = "".join([c for c in printables if c != ","])
         alphastring_ = OneOrMore(CharsNotIn(noncomma + ' '))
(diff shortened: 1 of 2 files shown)


johnyf:
Interesting: @kjellmf (the author of `dot2tex`) appears to have asked the `pyparsing` developer the exact same question about `QuoteString` here: http://pyparsing.wikispaces.com/share/view/3778969, in 2008.

johnyf:
The reason is that the module `dot2tex.dotparsing` is a borrowed version of the module `pydot.dot_parser`. The definition of `double_quoted_string` there has been [changed to](https://github.com/kjellmf/dot2tex/blob/75e634879a742995a407be154bf3ab73d06bf79a/dot2tex/dotparsing.py#L416):

```python
double_quoted_string = Regex(r'\"(?:\\\"|\\\\|[^"])*\"', re.MULTILINE)
```

which was the correction that the author of `dot2tex` used (the author of `pyparsing` suggested a regular expression that results in errors, and also notes that `QuotedString` is a little broken).

I do not know whether `QuotedString` has been fixed since 2008 (possibly not), and simple is better than complex (PEP 20), so I prefer the escape character that this PR proposes (which was the original suggestion of @kjellmf). I am not going to use workarounds to avoid illogical design in `pyparsing`; it is more reliable to instead stop using `pyparsing` in favor of `ply` in the future.

johnyf:
Merged as of 19156bdc393445e85b1caaf0110646be7ff244ab.

PR #135, opened 2016-10-06
BUG: fix networkx/networkx#2272

leycec:
Deserialization of graphs was critically broken under Python 2.x, resulting in deserialized graphs having no edges with no explicit warnings or errors. This pull request resolves both issues, parsing edges correctly and raising exceptions on unexpected edge cases for Python 2.x. <sup>_Get it, "edge"?_</sup>

In theory, pydot 1.2.3 should now be good to go. :+1:

### DIFF
--- a/dot_parser.py
+++ b/dot_parser.py
@@ -353,12 +353,13 @@ def push_edge_stmt(str, loc, toks):
 
         e.append(pydot.Edge(n_prev, name_port, **attrs))
 
-    elif isinstance(toks[2][0], type('')):
+    # Else if the target of this edge is the name of a node...
+    elif isinstance(toks[2][0], str_type):
 
         for n_next in [n for n in tuple(toks)[2::2]]:
 
             if (isinstance(n_next, P_AttrList) or
-                    not isinstance(n_next[0], type(''))):
+                    not isinstance(n_next[0], str_type)):
                 continue
 
             n_next_port = do_node_ports( n_next )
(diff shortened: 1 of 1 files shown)


johnyf:
Thanks for the bug fix. This bug originates from 842173cab55b57b0dbb753766edfa6c9d6842e87. I slightly modified the changes, mainly using `str.format` and raising `Exception` ([`SyntaxError` is raised by the Python parser](https://docs.python.org/2/library/exceptions.html#exceptions.SyntaxError)).

Regarding a unit test, I think that the change can be accepted as is. It would be good to add a test as [described here](https://github.com/erocarrera/pydot/issues/131#issuecomment-251877230) in the future.

johnyf:
Merged as 97061a8bf17d0ed205ff2c7fe92d389603d04558.

leycec:
> This bug originates from 842173c.

Feb 14, 2008 – over **six years** ago. Oh, Jeebus.

> ...using `str.format`

_Perfect._ I contribute to a few projects that prefer `%` (e.g., PyInstaller), but prefer `str.format` myself. Python 3.6 even introduces a new type of string interpolator: the [f-string](https://www.python.org/dev/peps/pep-0498).

Because apparently we don't have enough already.

> ...raising `Exception` (`SyntaxError` is raised by the Python parser).

Absolutely right. Best to always avoid type ambiguity.

> I think that the change can be accepted as is.

_And the wood bears danced with glee!_ :bear:

PR #137, opened 2016-10-12
Make tests always know their location.

prmtl:
This changes allow to run tests from any location without
requirement to changing dir to tests dir.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,4 @@ install:
   - pip install -r requirements.txt
 
 script:
-  - cd test/
-  - python pydot_unittest.py
-
-# after_success:
-#   - coveralls
+  - python test/pydot_unittest.py
--- a/test/pydot_unittest.py
+++ b/test/pydot_unittest.py
@@ -20,7 +20,7 @@ import unittest
 
 
 DOT_BINARY_PATH         = 'dot'
-TEST_DIR                = './'
+TEST_DIR                = os.path.dirname(__file__)
 REGRESSION_TESTS_DIR    = os.path.join(TEST_DIR, 'graphs')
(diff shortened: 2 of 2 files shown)


johnyf:
I did not write this part of the tests code. The code suggests automation, however I wouldn't have written it this way in the first place.

Following PEP20, I would choose the simpler, explicit, and flat approach. The reasons are:

1. These are tests, so the simpler they remain ([Kernighan's lever](http://www.linusakesson.net/programming/kernighans-lever/)), the fewer things that can go wrong
2. Nesting behavior within tests make debugging more difficult (this is a general comment, not closely relevant to this case)
3. is this automation necessary? Do the tests need to be run from another directory?
4. What is `__file__`?

Regarding running tests from the directory that has `pydot` underneath it, the wrong code will run. The reason is that `import pydot` within the tests module will be a relative import, and so not import the `pydot` installed under `site-packages`. However, it is the latter that we want to test, not the code inside the repository.

This should not be an issue with Python 3, or with [`from __future__ import absolute_import`](https://docs.python.org/2/library/__future__.html), or with a suitable `nose` configuration (in other projects). [Past experience](https://github.com/tulip-control/tulip-control/pull/145/commits/a827613cf277996641992cefa718a0c568ce876a) suggests that one can spend an afternoon tracking this flavor of simple but unexpected error, so I have grown conservative in this area.

About how this change behaves, i.e., what `__file__` is, in Python 2 it is not always an absolute path, as discussed [here](http://stackoverflow.com/a/7116925/1959808). Whether absolute or relative depends on whether `sys.path` contains the current directory. As a result, the tests will likely not run correctly in all environments.

prmtl:
To be honest I'm surprised about this probles with `__file__` (many thanks for such good explanation!). In this case it would definitely keep this as they are as this change would only add a (very) little gain and could produce a lot of hard to debug problems.

PR #138, opened 2016-10-12
Remove ResourceWarining from tests

prmtl:
Regression tests were raising ResourceWarning: unclosed file.


johnyf:
Good catch, thanks!

Issue #142, opened 2016-12-24
Cannot specify text encoding when writing files


brlam:
Currently the write function writes raw files without specifying a text encoding. This results in it using the default encoding of the system which is not always compatible with Graphviz. Since creating and writing other formats first write a raw temp file, this affects all formats.

Repro steps:
- Run Python in an environment where `locale.getpreferredencoding()` returns an encoding that is not supported by Graphviz. I used Windows in a Japanese locale which uses encoding "cp932"
- Run this code:
```python
import pydot
g = pydot.Dot()
n = pydot.Node('´') # Acute accent, not an ASCII char
g.add_node(n)
g.write_svg('test.svg')
```
- Open the created SVG file in Chrome and you will see an encoding error since it was output in the wrong encoding instead of UTF-8.

johnyf:
Thanks for pointing this out. After:

```
sudo apt-get install locales-all
sudo dpkg-reconfigure locales
```

on a Debian 9.1 (stretch), with:

```python
import locale

import pydot


locale.setlocale(locale.LC_ALL, 'lt_LT')
g = pydot.Dot()
n = pydot.Node('´')
g.add_node(n)
g.write_svg('test.svg')
```

I got the error:

```python
UnicodeEncodeError: 'charmap' codec can't encode character '\xb4' in position 13: character maps to <undefined>
```

PR #143, opened 2016-12-24
Add an encoding parameter to write_* and create_* functions

brlam:
Add an encoding parameter to `write_*` and `create_*` functions so that dot files created will be written in the specified encoding. This includes the temporary dot files created when writing or outputting in other formats.

Fixes issue #142.


johnyf:
Thanks, merged as 7e2f17644a954aae2412b6cb87a02fded7072745.

PR #150, opened 2017-03-22
Quote when necessary also for clusters

lgerardSRI:
It is odd (and a problem for my usage) that cluster names are not quoted when necessary while it is done for graphs, node, edge, etc.

### DIFF
--- a/pydot.py
+++ b/pydot.py
@@ -1635,7 +1635,7 @@ class Cluster(Graph):
         if obj_dict is None:
 
             self.obj_dict['type'] = 'subgraph'
-            self.obj_dict['name'] = 'cluster_'+graph_name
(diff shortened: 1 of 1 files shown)


johnyf:
Merged as b487c18c51a1842eab006ab4b969519f9cb5de3b, thanks for catching this issue.

Indeed:

- within `Node.__init__`: https://github.com/erocarrera/pydot/blob/35a8d858bd9da0b37268fe9b317fe4895387e75f/pydot.py#L569

- within `Edge.__init__`: https://github.com/erocarrera/pydot/blob/35a8d858bd9da0b37268fe9b317fe4895387e75f/pydot.py#L698-L699

- within `Graph.__init__`: https://github.com/erocarrera/pydot/blob/35a8d858bd9da0b37268fe9b317fe4895387e75f/pydot.py#L901

PR #152, opened 2017-04-26
Changed exception class to be more descriptive for graphviz installation errors

pnb:
pydot raises a generic `Exception` when it can't find graphviz installation binaries, which is not ideal. This pull request changes that to a more descriptive and specific class of exception so that it can be handled properly by other code (see https://github.com/fchollet/keras/pull/6398 for the specific motivation).

### DIFF
--- a/pydot.py
+++ b/pydot.py
@@ -1878,9 +1878,9 @@ class Dot(Graph):
                 stderr=subprocess.PIPE, stdout=subprocess.PIPE)
         except OSError as e:
             if e.errno == os.errno.ENOENT:
-                raise FileNotFoundError(
-                    '"{prog}" not found in path.'.format(
-                        prog=prog))
+                e.args = e.args + (
+                    '"{prog}" not found in path.'.format(prog=prog),)
+                raise OSError(*e.args)
             else:
                 raise
         stdout_data, stderr_data = p.communicate()
--- a/pydot.py
+++ b/pydot.py
@@ -1878,7 +1878,7 @@ class Dot(Graph):
                 stderr=subprocess.PIPE, stdout=subprocess.PIPE)
         except OSError as e:
             if e.errno == os.errno.ENOENT:
-                raise Exception(
+                raise FileNotFoundError(
                     '"{prog}" not found in path.'.format(
                         prog=prog))
             else:


johnyf:
Thank you for the suggestion. An issue with the selected exception `FileNotFoundError` is it [appears to be absent from Python 2.7](https://docs.python.org/2.7/library/exceptions.html). It is [present in Python 3.6](https://docs.python.org/3.6/library/exceptions.html#FileNotFoundError). `pydot` [still](https://github.com/erocarrera/pydot/blob/69dce474f6fbcaa6055586611453e43f3f20230d/setup.py#L28) declares compatibility with Python 2.7, so it cannot use an exception that is present only in Python 3.

pnb:
Good point! It's been a while since I used Python 2.7. Do you think ImportError would be appropriate in this case then?

johnyf:
`ImportError` is defined as ["Raised when the `import` statement has troubles trying to load a module. ..."](https://docs.python.org/3.7/library/exceptions.html#ImportError), but the file not found is a binary executable, not a python module.

pnb:
It's not an exact match for sure. Perhaps raising another OSError with the specific message would be appropriate then?

johnyf:
Raising `OSError` with additional information seems to be a good solution, because it is minimally invasive. It refines the current approach, because `OSError` is more specific than `Exception`. In intercepting `OSError` and replacing it with `Exception`, the current approach destroys information that can be preserved.

You may want to look at [this answer](http://stackoverflow.com/a/9157277/1959808), so that the exact same error be raised, except for a modification to its message. This would allow higher levels (e.g., `keras`) to handle a very specific exception type as if `pydot` was transparent, by checking themselves whether `e.errno == os.errno.ENOENT`.

pnb:
I think that will work great! The approach in that stack overflow answer didn't seem to work across both python 2 and 3, as `e.args` was not re-interpreted when the exception was thrown for some reason. However it was easy to raise a new `OSError` with all the old arguments plus a new descriptive argument.

In python 3 apparently this is automatically turned into `FileNotFoundError`, still with the custom message, which seems like a pretty ideal outcome.

johnyf:
Merged as edb07209cc1c90ddab1a5033a03ce1c64b8388cc.

Thanks for adjusting this change. Regarding the observation that in Python 3 instantiating [`OSError`](https://docs.python.org/3/library/exceptions.html#OSError) returns the subclass [`FileNotFoundError`](https://docs.python.org/3/library/exceptions.html#FileNotFoundError):

> The constructor often actually returns a subclass of `OSError`, as described in OS exceptions below. The particular subclass depends on the final `errno` value. This behaviour only occurs when constructing `OSError` directly or via an alias, and is not inherited when subclassing.

In particular, `FileNotFoundError`:

> Corresponds to `errno ENOENT`.

Also, appending the message at the end of `OSError.args` didn't match the signature of its constructor. Instead,  modifies only `strerror` and `filename`. Tested with both Python 2 and 3.

PR #153, opened 2017-05-02
Propagate LD_LIBRARY_PATH when calling graphviz

chrisburr:
Since [v1.2.0](https://github.com/erocarrera/pydot/commit/bc639e76b214b1795ebd6263680ee55d9d4fca9f#diff-44fda7721399b25e8cb06471016a3454L1942) `pydot` fails when used with a installation of graphviz which requires `LD_LIBRARY_PATH` to be set appropriately. This PR re-enables the propagation of `LD_LIBRARY_PATH` to the subprocess.

I'm not sure if any other variables should be included? The other main one that I can think of is `DYLD_LIBRARY_PATH`.

### DIFF
--- a/pydot.py
+++ b/pydot.py
@@ -1868,6 +1868,7 @@ class Dot(Graph):
         # with `shell=False`
         env = dict()
         env['PATH'] = os.environ['PATH']
+        env['LD_LIBRARY_PATH'] = os.environ.get('LD_LIBRARY_PATH', '')
         cmdline = [prog, '-T' + format] + args + [tmp_name]
         try:
             p = subprocess.Popen(


johnyf:
Merged as ca7fc910fb161cca7ca61e168bc40a3e1971f7db.

This is a good point.

The proposed change raises also another issue: that in principle there may be no environment variable `PATH` defined (a theoretical scenario would be to have a symbolic link to `dot` within the current directory). An `unset PATH` suffices to demonstrate the `KeyError` raised in this use case. Addressed in 926d8b2eddf40c07e8a187c3c61e95750e5db14e.

I think that `LD_LIBRARY_PATH` suffices (unless anyone explains why it would be necessary). `DYLD_LIBRARY_PATH` seems to have [more complicated behavior](https://stackoverflow.com/questions/3146274/is-it-ok-to-use-dyld-library-path-on-mac-os-x-and-whats-the-dynamic-library-s) across different operating systems, so simpler is better than complex [PEP 20].

PR #156, opened 2017-08-24
- use utf8 default encoding

mahomahomaho:
Pydot was throwing unicoencode error ('ascii' codec) before that change, when I was using Polish national characters in labels but server was running in LC_ALL=C environment.


johnyf:
Thanks for the proposed change. It is similar to #143, which motivates more this API change.

johnyf:
The changes in #143, which has been merged, cover this behavior.

Issue #145, opened 2017-01-04
Cannot install with PYTHONOPTIMIZE=2


Yajo:
Steps to reproduce:

```
$ docker run -e PYTHONOPTIMIZE=2 -it --rm python:2-alpine pip install pydot
Collecting pydot
  Downloading pydot-1.2.3.tar.gz
    Complete output from command python setup.py egg_info:
    pydot.py:33: UserWarning: Couldn't import dot_parser, loading of dot files will not be possible.
      "Couldn't import dot_parser, "
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-eqA_Ct/pydot/setup.py", line 32, in <module>
        long_description="\n".join(pydot.__doc__.split('\n')),
    AttributeError: 'NoneType' object has no attribute 'split'
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-eqA_Ct/pydot/
```

When using [PYTHONOPTIMIZE=2](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONOPTIMIZE), [`__doc__` are removed from the generated `*.pyo` files](https://docs.python.org/3/using/cmdline.html#cmdoption-OO), assuming that in a production deployment you never need the docstrings.

Could you please use an alternative method to get that docstring and make it work with maximum optimization?

Thanks!

johnyf:
Thank you for noting this behavior. The plan is to load the long description from the README file, so it should address this problem by not accessing a module docstring. However, there are situations (not necessarily within `pydot`) where docstrings are necessary, for example for parser generators that use `ply`.

Yajo:
But do you mean in a production environment?

johnyf:
To be more accurate, [`ply`](https://github.com/dabeaz/ply) reads production rules defined within docstrings and produces a parser. It can save parser tables for later use, as an optimization. I am not sure whether `ply` can be configured to use those tables directly, without the code that defines the production rules.

In any case, `ply` is not used here (it could be in the future, if `pyparsing` was replaced, see https://github.com/erocarrera/pydot/issues/132#issuecomment-250311252). I mentioned this case to suggest that there are use cases where docstrings would have a function within a production environment.

Yajo:
Oh of course they have. That's why we have `PYTHONOPTIMIZE=1` too 😉.

[...]

PR #160, opened 2017-10-08
Load package description from README

prmtl:
To have a good formating of package description,
it was rewritten from Markdown to reStructuredText

Fixes: #145

### DIFF
--- a/pydot.py
+++ b/pydot.py
@@ -1,20 +1,3 @@
-"""Graphviz's dot language Python interface.
-
-This module provides with a full interface to
-create handle modify and process graphs in
-Graphviz's dot language.
-
-References:
-
-pydot Homepage: https://github.com/erocarrera/pydot
-Graphviz:       http://www.graphviz.org/
-DOT Language:   http://www.graphviz.org/doc/info/lang.html
-
-Copyright (c) 2005-2011 Ero Carrera <ero.carrera@gmail.com>
-
-Distributed under MIT license
-[http://opensource.org/licenses/mit-license.html].
-"""
 from __future__ import division
 from __future__ import print_function
 import copy
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,22 @@
 #!/usr/bin/env python
-"""Installation script."""
+"""Graphviz's dot language Python interface.
+
+This module provides with a full interface to
+create handle modify and process graphs in
+Graphviz's dot language.
+
+References:
+
+* pydot Homepage: https://github.com/erocarrera/pydot
+* Graphviz:       http://www.graphviz.org/
+* DOT Language:   http://www.graphviz.org/doc/info/lang.html
+
+Copyright (c) 2005-2011 Ero Carrera <ero.carrera@gmail.com>
+
+Distributed under MIT license
+[http://opensource.org/licenses/mit-license.html].
+"""
+
 try:
     from setuptools import setup
 except ImportError:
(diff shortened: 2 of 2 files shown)


prmtl:
Preview of new description: https://test.pypi.org/project/test-of-pytdot/1.2.6/

johnyf:
Thanks. However, I prefer Markdown, because it is easier to write and read in source form (e.g., there is a typo in the preview: "usin``g"). The [current long description on PyPI](https://pypi.org/project/pydot/1.2.3/) is not from the README, so it has good formatting. GitHub displays the Markdown README correctly.

Regarding #145, I now favor a simpler approach: to place the package's docstring within the script `setup.py`, as a short paragraph. Examples of packages that follow this simple approach are `networkx`, `numpy`, `scipy`, and `matplotlib`. Their "long" descriptions range from 1 line to a short paragraph. `pydot` follows the same approach, it is just that the description happened to be loaded from a module's docstring. This approach is independent of formatting, and encourages the reader to visit the repository in order to read the detailed README (which can be in Markdown).

Besides, I think that `warehouse` and other distribution infrastructure will eventually support Markdown, https://github.com/pypa/warehouse/issues/869.

prmtl:
Fixed the typo.
I'm open to do it your way (just want #145 to be fixed), I just wanted to share my point with you:

>The current long description on PyPI is not from the README, so it has good formatting.

No, it doesn't. Line breaks are missing.

> GitHub displays the Markdown README correctly.

Just like ReST: https://github.com/prmtl/pydot/tree/fix/pythonoptimize_issue_145

> Regarding #145, I now favor a simpler approach: to place the package's docstring within the script setup.py, as a short paragraph. Examples of packages that follow this simple approach are networkx, numpy, scipy, and matplotlib. Their "long" descriptions range from 1 line to a short paragraph. pydot follows the same approach, it is just that the description happened to be loaded from a module's docstring. This approach is independent of formatting, and encourages the reader to visit the repository in order to read the detailed README (which can be in Markdown).

And then it is a bunch of packages (`httpie`, `requests`) that use README for this purpose. The plus is that you have just one source of package description. Also rich format like ReST or Markdown gives a nice way to include links, images, etc. into description on PyPI.

> Besides, I think that warehouse and other distribution infrastructure will eventually support Markdown, pypa/warehouse#869.

It seems that this will not happen in near future.

johnyf:
Good point about the missing line breaks.

I didn't intend to suggest that GitHub doesn't render ReST, but that it does render Markdown, so that it currently works.

If the package description on PyPI is short, then it doesn't duplicate the information in the README. It is a high-level summary that remains usually unchanged between releases. In other words, the long description on PyPI doesn't need to be regarded as the README, it can be just an abstract.

Regarding the choice between ReST and Markdown, I find Markdown easier to write. This is a personal preference; others may find ReST easier.

prmtl:
Updated.

johnyf:
Thanks, but with `PYTHONOPTIMIZE=2` (#145), this change passes `None` as `long_description` to `setuptools.setup`. This can be observed using `print(__doc__)` inside `setup.py`.

Regarding PR branches, please avoid force pushing, as also [recommended for CPython](https://docs.python.org/devguide/pullrequest.html#submitting). Rebasing will be performed at the time of merging.

Other comments:

- The module `pydot` needs some docstring.
- The package's description does not describe the script `setup.py`.
- More changes needed to fully address #145 (f374f66b93ec7e6bb7a6812abdb2d8a62db879f8), i.e., behave compatibly with `PYTHONOPTIMIZE=2`.
- Mentioning issues from commits works better only after changes (e.g., in a PR) have been finalized. Otherwise it creates multiple copies of the same reference in the mentioned issue (when force pushing or amending), and also GitHub will retain those commits in the future, even if they were on some branch or fork and have been rebased or replaced. I have come to prefer documenting within issues which commits addressed them, after merging those to `master`.

Merged as of f374f66b93ec7e6bb7a6812abdb2d8a62db879f8

johnyf:
[PyPI now supports Markdown READMEs](https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi) (as anticipated [above](https://github.com/erocarrera/pydot/pull/160#issuecomment-335030514)).

PR #181, opened 2018-07-11
Fixes Issue180 -- get_node() get_edge() error when name was quoted

### LINKED ISSUE #180: get_node() and get_edge() does not match correctly if the original name was quoted
To replicate the error:

```python
graph = pydot.Dot()
nodeA = pydot.Node('Node A')
graph.get_node('Node A')  # This return an empty list, whereas it should return [nodeA]
```

This error happens when the original names (given by the user) and the quoted names (used in inner dicts as keys) are mismatched.

qinyeli:
Hello.
This pull request fixes [Issue 180](https://github.com/erocarrera/pydot/issues/180)

### DIFF
--- a/pydot.py
+++ b/pydot.py
@@ -1139,6 +1139,8 @@ class Graph(Common):
         An empty list is returned otherwise.
         """
 
+        name = quote_if_necessary(name)
+
         match = list()
 
         if name in self.obj_dict['nodes']:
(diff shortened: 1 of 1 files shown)


peternowee:
See my reply in https://github.com/pydot/pydot/issues/180#issuecomment-735272000.

PR #186, opened 2018-10-14
CI: Replace 3.7-dev with 3.7 final

hugovk:
`3.7-dev` is an old alpha version (`3.7.0a4+`) that hasn't been updated for a long time.

Python 3.7.0 final was released on 2018-06-27, so let's use that instead.

It needs a little workaround on Travis, see https://github.com/travis-ci/travis-ci/issues/9815.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,13 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
-  - "3.7-dev"
+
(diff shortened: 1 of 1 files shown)


hugovk:
The build failure of Python 2.6 is unrelated, and will be fixed by https://github.com/erocarrera/pydot/pull/185.

PR #194, opened 2018-11-25
Add support of Windows' conda installations

prmtl:
On conda/anaconda Graphviz is called using .bat files due to some
customizations made on that platform [1].

[1] https://github.com/conda-forge/graphviz-feedstock/blob/24a27a82edc23ea0581f9ee5b2c757fba66907b1/recipe/bld.bat#L26)


peternowee:
See pydot/pydot-ng#52 for earlier discussion that led to this change.

PR #198, opened 2018-12-05
Fixing failing regression tests

prmtl:


prmtl:
Obsolete in favor of #211

Issue #203, opened 2018-12-14
Problem with graph.write.png...AssertionError: 1


marcorusc:
So i'm running this code and it tells me that there is a certain syntax error in /tmp/tmpgIPxNG
Something is not working when i tried to generate the graph... this is the error i have:
-------------------------------------------------------------------------------------------------------------------------------
/usr/bin/python2.7 /home/marco/PycharmProjects/AIforvideogame/Treeplottry.py
"dot" with args ['-Tpng', '/tmp/tmpgIPxNG'] returned code: 1

stdout, stderr:
 
Traceback (most recent call last):
Error: /tmp/tmpgIPxNG: syntax error in line 2 near '>'

  File "/home/marco/PycharmProjects/AIforvideogame/Treeplottry.py", line 23, in <module>

    treeplot.generateDiagram(root, childNode)
  File "/home/marco/PycharmProjects/AIforvideogame/Treeplot.py", line 58, in generateDiagram
    self.graph.write_png('graph.png')
  File "/home/marco/.local/lib/python2.7/site-packages/pydot.py", line 1734, in new_method
    encoding=encoding)
  File "/home/marco/.local/lib/python2.7/site-packages/pydot.py", line 1817, in write
    s = self.create(prog, format, encoding=encoding)
  File "/home/marco/.local/lib/python2.7/site-packages/pydot.py", line 1945, in create
    assert process.returncode == 0, process.returncode
AssertionError: 1

Process finished with exit code 1
-----------------------------------------------------------------------------------------------------------------
I wrote this method:
---------------------------------------------------------------------------------------------------------------------
    def generateDiagram(self, rootNode, currentNode):

        # this method generetes diagram

        # add nodes to edges to graph
        self.createGraph(rootNode, currentNode)
-----------------------------------------------------------------------------------------------------
Honestly I don't know what to do anymore....

prmtl:
From the error, I cannot see more that there might be a problem with generated DOT string that is passed to `dot` program to render a graph.

It would be extremely helpful to see full code that generate graph or the resulting DOT string (it can be checked by printing results of `some_graph.to_string()`)

marcorusc:
> From the error, I cannot see more that there might be a problem with generated DOT string that is passed to `dot` program to render a graph.
> 
Ok thanks for the help! here is the code:
 def generateDiagram(self, rootNode, currentNode):

        # this method generetes diagram

        # add nodes to edges to graph
        self.createGraph(rootNode, currentNode)

[...]

Issue #171, opened 2018-04-26
Exception on parsing dot file


goerz:
With `structure.dot` as

    digraph qnet {
        rankdir="TB";
        graph [pad="0", ranksep="0.25", nodesep="0.25"];
        node [penwidth=0.5, height=0.25, color=black, shape=box, fontsize=10, fontname="Vera Sans, DejaVu Sans, Liberation Sans, Arial, Helvetica, sans"];
        edge [penwidth=0.5, arrowsize=0.5];
        compound=true;

        subgraph cluster_qnet {
            fontname="Vera Sans, DejaVu Sans, Liberation Sans, Arial, Helvetica, sans";
            fontsize=12;
            label="qnet";
            tooltip="qnet";

            subgraph cluster_algebra {
                toolbox          [target="_top"; tooltip="qnet.algebra.toolbox",          href="../API/qnet.algebra.toolbox.html"];
                library          [target="_top"; tooltip="qnet.algebra.library",          href="../API/qnet.algebra.library.html"];
                core             [target="_top"; tooltip="qnet.algebra.core",             href="../API/qnet.algebra.core.html";             width=1.3];
                pattern_matching [target="_top"; tooltip="qnet.algebra.pattern_matching", href="../API/qnet.algebra.pattern_matching.html"; width=1.3];
                toolbox -> library;
                toolbox -> core;
                toolbox -> pattern_matching;
                library -> core;
                core    -> pattern_matching [weight=0, minlen=0];
                library -> pattern_matching;
                href = "../API/qnet.algebra.html"; target="_top";
                label="algebra";
                tooltip="qnet.algebra";
                graph[style=filled; fillcolor="#EEEEEE"];
            }

            convert          [target="_top"; tooltip="qnet.convert",       href="../API/qnet.convert.html"];
            visualization    [target="_top"; tooltip="qnet.visualization", href="../API/qnet.visualization.html"];
            printing         [target="_top"; tooltip="qnet.printing",      href="../API/qnet.printing.html"];
            utils            [target="_top"; tooltip="qnet.utils",         href="../API/qnet.utils.html"; width=1];

            { rank=same; convert[width=0.8]; visualization[width=0.8]; printing[width=0.8]; }
            convert       -> visualization [minlen=3, style=invis];
            visualization -> printing      [minlen=3];
            visualization -> toolbox  [minlen=2, lhead=cluster_algebra];
            printing      -> toolbox  [lhead=cluster_algebra];
[...]

PR #218, opened 2019-10-05
[WIP] Improve our handling of Graphviz errors

peternowee:
When pydot calls Graphviz and Graphviz returns a non-zero exit code (indicating an error), pydot currently reports the Graphviz stdout and stderr output and raises an exception.

This commit adds to that error report the 'raw' DOT string generated by pydot, as it was also sent to Graphviz as input.

Also, this commit separates the Graphviz stdout and stderr output more clearly.

These changes should help in troubleshooting Graphviz errors.

For some examples where this would have helped see: pydot/pydot#203 ("Problem with graph.write.png...AssertionError: 1")

Tested with Python 2.7 and 3.7. All tests pass (when run together with the test suite fix proposed in PR 211).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,12 @@
 # `pydot` changelog
 
 
+1.4.2 (unreleased)
+------------------
+
+- Nothing changed yet.
+
+
 1.4.1 (2018-12-12)
 ------------------
 
--- a/pydot.py
+++ b/pydot.py
@@ -20,7 +20,7 @@ except Exception as e:
 
 
 __author__ = 'Ero Carrera'
-__version__ = '1.4.1'
+__version__ = '1.4.2.dev0'
 __license__ = 'MIT'
 
 


prmtl:
Hey! I just wanted to let you (and everyone interested) that this repo is not abandoned. I know that it was a long time without attention from my side, but I am going to look at all new issues/comments/PRs in a few days.

peternowee:
:D Great, good to know, I was starting to wonder already! Thanks for letting us know!

peternowee:
Just let me know if you want any changes, also in commit message, or if I need to rebase a branch or merge master, whatever you prefer.

prmtl:
I like the idea, the only issue I see is that even with smaller graphs it makes errors harder to read, not to mention bigger ones.
I would rather see it as an option (kind of a "debug" or "verbose" flag/option?) that user would have to explicitly use.

peternowee:
Sorry, now it's my turn to apologize for the delay. Life gets in the way sometimes. This will continue for at least a few more weeks, so let's try to take smaller steps.

I agree that printing long error messages is not desirable. In fact, I found that any printing of messages is not very desirable for a library and I had been considering getting rid of printing altogether and use exceptions instead. All the debug output can be attached to a custom exception as attributes.

A problem common to both a debug/verbose flag and adding the output as attributes of an exception, is that both methods depend on the main program (the program interacting the user) and all the modules in the middle to help communicate between pydot and the end user. If they do not cooperate (e.g. the debug/verbose flag is not passed on to pydot, or the additional attributes of the pydot exception are not extracted and shown to the user by the main program), it would actually have become much more difficult for the average end user to provide the debug output than when it was simply printed. I am afraid that in those cases, we would just have to ask those end users to discuss their issue with the developers of the main program first.

If you want to explore the use of exceptions more, I have some further questions for you, but maybe you can first let me know what you think about the above.

sytelus:
Wondering if this project is dead. No commits for 1+ year and no PRs merged for LONG time. Time to fork this thing? Current release has stopped functioning in Windows 10.

peternowee:
@ankostis You [commented on commit 51d5dcdc in sytelus/pydotz](https://github.com/sytelus/pydotz/commit/51d5dcdc25c1dc5fdc4aceaa2f5f69099a2c3030#commitcomment-38571682) (a pydot-fork). That commit is part of this PR pydot/pydot#218, which is still open here in pydot, but was already merged in pydotz. My replies with regard to pydotz are below your comment there. My replies with regard to pydot follow here:

> Just to be clear, this PR is not adequate, according to our discussion in [pydot#219](https://github.com/pydot/pydot/pull/219).

Talking just about pydot, you are right that the outcomes of the current discussions on error handling in pydot/pydot#219 and pydot/pydot#171 will heavily influence this PR (pydot/pydot#218) as well. I raised the idea of using custom exceptions with debug output attached as attributes (similar to what is now discussed in PR 219 for another code segment) here in [pydot#218 in November 2019](https://github.com/pydot/pydot/pull/218#issuecomment-555360553) already. I am happy that you confirmed my feeling and all the input you gave me, but I really need some more time to digest that and read up on the subject a bit myself, before I can respond to some of it and make new proposals.

> It better raise, specifically a new Exception that stores the stdout/stdin in ts `*args` and formats the message out of them
> (instead of permanently grafting them into it).
> Is m understandin correct?

I am still reading, watching, learning, thinking and trying, but it seems that the use of exceptions will indeed be the way forward. I don't know about stdin, but in this case at least stdout, stderr, the DOT-file, the called program, arguments and exit code need to be stored. Maybe I am still missing something.

With "formats the message out of them (instead of permanently grafting them into it)" you mean that all of those elements are stored as a separate attributes of the exception object and that they are only merged into a long string when that is requested, right?

By the way, if you want to move faster than me, feel free to start working on your own PR. I really don't mind. If I like it, I will even close this one. I would appreciate it if you would let me know then, to prevent that we both spend time on the same thing.

Finally, I want to copy in a part [my comment of last week in #171](https://github.com/pydot/pydot/issues/171#issuecomment-612655762) about an `InvocationException` still lying around in the current code. If we re-purpose it, it should probably be used for this PR #218, not #219:

> > By the way, there is still an unused class lying around in the pydot code: `InvocationException`, in disuse since commits 9b3c1a1 and bc639e7 of 2016 (v1.2.0). I think it was actually meant to do something similar as we are discussing here, so maybe its removal in 2016 may actually be considered somewhat of a regression. Maybe we can revive/revert it, extend it and even make it backward compatible with its earlier behavior again. OTOH, we might need to distinguish two separate classes, one for errors from calls to external programs (Graphviz `dot`, issue 218) and one from errors from calls to Python libraries (pyparsing), because they will carry different context data.

ankostis:
> With "formats the message out of them (instead of permanently grafting them into it)" you mean that all of those elements are stored as a separate attributes of the exception object and that they are only merged into a long string when that is requested, right?

Exactly.

> I am still reading, watching, learning, thinking and trying, but it seems that the use of exceptions will indeed be the way forward. I don't know about stdin, but in this case at least stdout, stderr, the DOT-file or DOT-string

These should be enough. Sometimes the `cwd` of the executed process is relevant - but it might not be useful here where always `dot` with a single file is executed.

> By the way, if you want to move faster than me, feel free ...

Sorry, not much time available form me at the moment.
If i had it, i would re-use my code of class `os.cmd` i had linked a week ago.
Maybe it firs into `InvocationException` usecase.

peternowee:
Following the discussion in pydot/pydot#171, I have been working on better solutions for PRs pydot/pydot#218 and pydot/pydot#219. These solutions depend on dropping Python 2 support and having a logging facility. For this, I created PRs pydot/pydot#229 and pydot/pydot#231. Also, pydot/pydot#211 needs to be merged to make the Travis CI test results meaningful again.

I will now suspend my work until at least those 3 PRs (pydot/pydot#211, pydot/pydot#229, pydot/pydot#231) have been merged into the master branch. I remain available to rebase my PRs or advise about the order in which these and other PRs can best be merged.

peternowee:
Should be solved as part of pydot/pydot#218. That PR is currently closed, but I hope to reopen it in the future.

PR #219, opened 2019-10-06
[WIP] Improve our handling of pyparsing errors

### LINKED ISSUE #176: Error in error handling hides the real error message
https://github.com/erocarrera/pydot/blob/d6ac9e9244d1a882103422ac2b35ceef96f5dfe3/dot_parser.py#L554

probably want `str(err)`, not just `err`? Currently I am getting

```
pydot raised an error: must be str, not ParseException
```

When I would expect to be getting?

```
pydot raised an error: Expected "}" (at char 688), (line:19, col:29)
```

peternowee:
_From the commit message:_

When parsing DOT strings or files, pydot's parser calls the third-party library pyparsing. If pyparsing meets a problem, it raises a `ParseException`, which pydot's parser then reports.

This commit makes some improvements to that reporting:

- Use `str(err)` instead of just `err` to prevent errors such as:
  Python 2.7:

      TypeError: cannot concatenate 'str' and 'ParseException' objects

  Python 3.7:

      TypeError: can only concatenate str (not "ParseException") to str

  This closes https://github.com/pydot/pydot/issues/176 ("Error in error handling hides the real error message"), reported by Shish.
- Add newlines where necessary.
- When running under Python 3, use the `ParseException.explain()` method if it is available. This method was introduced in pyparsing 2.3.1 of January 2019 [1] [2]. It returns the same information that we otherwise manually string together (line, column arrow and error message), and adds to that a list of the pyparsing expressions that caused the exception to be raised.

I verified the new behavior by manually testing under the four combinations of Python 2.7.16/3.7.3 and pyparsing 2.3.0/2.4.2.

Unit tests all pass (same four combinations, when run together with the test suite fix proposed in PR 211).

[1]: https://github.com/pyparsing/pyparsing/blob/pyparsing_2.4.2/CHANGES#L396
[2]: https://github.com/pyparsing/pyparsing/blob/pyparsing_2.4.2/pyparsing.py#L378

_Additional remarks:_

Example output with Python 2.7.16 and pyparsing 2.4.2:

    >>> pydot.graph_from_dot_data("graph { n [label=<some complex problematic DOT string>>]; }")
    graph { n [label=<some complex problematic DOT string>>]; }
              ^
    Expected "}", found '['  (at char 10), (line:1, col:11)

Example output with Python 3.7.3 and pyparsing 2.4.2:

    >>> pydot.graph_from_dot_data("graph { n [label=<some complex problematic DOT string>>]; }")
    graph { n [label=<some complex problematic DOT string>>]; }
              ^
    ParseException: Expected "}", found '['  (at char 10), (line:1, col:11)
    parse_dot_data
    pyparsing.OneOrMore - {{['strict'] Group:({'graph' | 'digraph'}) [ID] graph_stmt}}...
    pyparsing.And - {['strict'] Group:({'graph' | 'digraph'}) [ID] graph_stmt}
    pyparsing.Group - graph_stmt
    pyparsing.And - {Suppress:("{") [Forward: {{stmt [Suppress:(";")]}}...] Suppress:("}") [Suppress:(";")]}
    pyparsing.Suppress - Suppress:("}")
    pyparsing._SingleCharLiteral - "}"

Let me know if I need to split this commit in two commits, the second commit then to introduce the use of `explain()`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,12 @@
 # `pydot` changelog
 
 
+1.4.2 (unreleased)
+------------------
+
+- Nothing changed yet.
+
+
 1.4.1 (2018-12-12)
 ------------------
 
--- a/pydot.py
+++ b/pydot.py
@@ -20,7 +20,7 @@ except Exception as e:
 
 
 __author__ = 'Ero Carrera'
-__version__ = '1.4.1'
+__version__ = '1.4.2.dev0'
 __license__ = 'MIT'
 
 


peternowee:
For an alternative solution by @ankostis and discussion about bubbling-up and/or printing details of the `ParseException`, see https://github.com/pydot/pydot/issues/171#issuecomment-612397349 and later comments.

dot_parser.py
```diff
@@ -548,8 +548,13 @@ def parse_dot_data(s):
         tokens = graphparser.parseString(s)
         return list(tokens)
     except ParseException as err:
-        print(
-            err.line +
-            " "*(err.column-1) + "^" +
-            err)
+        if PY3 and (hasattr(ParseException, 'explain') and
```
ankostis, line 551: It is recommended to always check if PY2 (in case PY4 ever comes out :-)

dot_parser.py
```diff
@@ -548,8 +548,13 @@ def parse_dot_data(s):
...
+                    callable(getattr(ParseException, 'explain'))):
+            print(ParseException.explain(err))
```
ankostis, line 553: Shouldn't it print to `stderr`, or use `logging` standard-module?

dot_parser.py
```diff
@@ -548,8 +548,13 @@ def parse_dot_data(s):
...
+        else:
+            print(
```
ankostis, line 555: See comment about `stderr`, `logging`, above.

dot_parser.py
```diff
@@ -548,8 +548,13 @@ def parse_dot_data(s):
...
+                err.line + "\n" +
+                " "*(err.column-1) + "^\n" +
+                str(err)
```
ankostis, line 558: Isn't it better to use format-strings (or `logging` does that).

dot_parser.py
```diff
@@ -548,8 +548,13 @@ def parse_dot_data(s):
         tokens = graphparser.parseString(s)
         return list(tokens)
     except ParseException as err:
-        print(
-            err.line +
-            " "*(err.column-1) + "^" +
-            err)
+        if PY3 and (hasattr(ParseException, 'explain') and
+                    callable(getattr(ParseException, 'explain'))):
```
ankostis, line 552: What does `explain` attribute contain from (stdout, stderr, cmd-line, cmd-args)?

Would it worth it to make it configurable with a package-level configuration flag, e.g `popen_ex_contets` with bit-map values: 
- `POPEN_EX_INCLUDE_ALL`
- `POPEN_EX_INCLUDE_STDOUT`
- `POPEN_EX_INCLUDE_STDERR`
- `POPEN_EX_INCLUDE__CMD_LINE`
- `POPEN_EX_INCLUDE__CMD_ARGS`

to allow for client code mitigate any privacy concerns iof the dot content
and make a new `DEBUG` flag translate to (see my comment at the bottom):
```python
if DEBUG:
    popen_ex_contets = POPEN_EX_INCLUDE_ALL
    raise_errors = True
```

dot_parser.py
```diff
@@ -548,8 +548,13 @@ def parse_dot_data(s):
...
+            print(ParseException.explain(err))
+        else:
+            print(
+                err.line + "\n" +
+                " "*(err.column-1) + "^\n" +
+                str(err)
+                )
         return None
```
ankostis, line 560: Wouldn't be better to add a new package-level config-option (e.g. `pydot.raise_errors`), not to ever return `None`,
but always let exceptions bubble-up?

ankostis (commented):
I distributed my comments on #219 in this review about enhancing pysot error-handling of `Popen` errors..

dot_parser.py
peternowee: Hi @ankostis, thanks for reviewing my proposed changes, I really appreciate it. I will go by them one by one, as I find time.

About checking `PY3`. I am not sure if you are aware of the following:
- `PY3` is already used elsewhere in the code: 3 conditionals in `pydot.py` and 1 conditional in `dot_parser.py` use it already.
- Both files set it like this: `PY3 = sys.version_info >= (3, 0, 0)`, meaning Python 3 or higher. Maybe you were assuming that it was defined as in the [`six` library](https://github.com/benjaminp/six/blob/1.14.0/six.py#L37), as `sys.version_info[0] == 3`, meaning Python 3 _only_, in which case I would have agreed with your recommendation, but it is not.

You could still say:
- That the pydot definition of `PY3` is different than elsewhere and therefore confusing for pydot developers. In reply, I would say that I expect most developers will look up the `PY3` definition in the code when they run into it (as I did) and that any mistakes that result from any confusion may still get caught through code review or automated testing as well.
- That it might become more confusing once Python 4 is out, because people might then start to use `PY3` to mean 'not on Python 4 yet', which actually cannot be determined using our definition of `PY3`. In reply, I would say that the world may look very different by the time Python 4 comes out. The affected code lines may have been changed a dozen times between now and then already. Maybe they will have been deleted a long time before already, during a drop of support for Python 2. And again, I believe that the developers of that time will read the definition in the source code and change it as they see fit when adding conditionals for Python 3+4 support.

In total, I am not really against changing to `PY2`, but I also do not really see much benefit. In case you still recommend checking for `PY2`, can you at least agree that it should be done separate from this PR? Because it will require carefully changing 2 variable settings and flipping 4 conditionals (5 if you count this one) and that is all not really related to the subject of this PR.

ankostis: > That it might become more confusing once Python 4 is ...

Guido was clear on that, there won't ever be Python 4 (but otoh, Guido is not BDFL anymore:-) 

The whole point of checking PY2 (instead of PY3) is to sidestep our mental model of what it means a "regular python" - PY2 is the exception now.
In principle, you should be able to delete all PY2 conditionals, and the code should still work flawlessly in a modern OS.

In any case, i would always stay close to the official recommended procedures for porting PY2 code: https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection

peternowee: > The whole point of checking PY2 (instead of PY3) is to sidestep our mental model of what it means a "regular python" - PY2 is the exception now.

Ok, I see, valid point. Still I hesitate to make that change in this PR now.

> In principle, you should be able to delete all PY2 conditionals, and the code should still work flawlessly in a modern OS.

Agreed, and this may be the easiest way out of this discussion. I am hoping there will be at least one more point release in the pydot 1.4.x branch that still supports Python 2, to provide those users with some of the important bug fixes that are currently awaiting approval. And then drop Python 2 and <3.5 support in a pydot 1.5.x version. Indeed that would clean up a lot of code and make life a lot easier. Parallel to 1.5.x, we could keep the 1.4.x branch alive for Python 2 users, backporting bug fixes as long as there are people willing to work on that.

> In any case, i would always stay close to the official recommended procedures for porting PY2 code: https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection

Yes, I had read that section as well, but I still thought pydot's `PY3` definition was not so bad, at least not as bad as the first example.

peternowee: Maybe I should split this PR into the fix for the `TypeError` and the change/new 'feature' of using `ParseException.explain()`. That new feature would not fit in a point release anyway, so it would go to 1.5.x, so no need for `PY3` then anymore anyway. The `TypeError` fix is simple and can go in both branches 1.4.x and 1.5.x.

dot_parser.py
peternowee: Ok, after some more reading, I certainly agree with you that `None` should not be returned.

Some sources:
- [MIT 6.0001 Introduction to Computer Science and Programming in Python, Fall 2016 by Dr. Ana Bell](https://www.youtube.com/watch?v=9H6muyZjms0&t=1678). On handling exceptions by substituting with a default value: "[T]he user thinks that they entered something, and they think everything's great, your program accepts it, but then they get some weird value as an output, which is far from what they expected. So it's not really a good idea to just replace users' values with anything."
- https://stackoverflow.com/questions/1630706/best-practice-in-python-for-return-value-on-error-vs-success
- https://github.com/williballenthin/python-registry/issues/76
- https://wiki.python.org/moin/HandlingExceptions

In our discussion in issue [#171](https://github.com/pydot/pydot/issues/171#issuecomment-612582791), I worried that some users may depend on a possible return value `None`. In other words, removing that may be a breaking change. Some remarks about that:
- I checked the current pydot 1.4.1 code and did not see anything directly or indirectly depending on a return value of `None` from this function. I also checked [NetworkX 2.4](https://github.com/networkx/networkx/tree/networkx-2.4) code as an example of a library that depends on pydot and also did not find any there. (There is of course a multitude of dependent code out there that I did not check.)
- The current docstring of the function states that it returns Graphs and does not mention possibly returning `None`.
- Still, I think changes in return values (and possible exceptions) are too far-reaching for a mere point release (e.g. 1.4.2). So, I will split off the bug fix for the `TypeError` to a separate PR (as soon as we agree on the solution for that). This PR (219) will keep the more substantial changes, including the removal of `return None`, for the next major or minor release (e.g. 1.5.x). I will add a Changelog-entry and any necessary function docstring changes to this PR in an effort to make users aware of the change.

> Wouldn't be better to add a new package-level config-option (e.g. `pydot.raise_errors`), not to ever return `None`, but always let exceptions bubble-up?

If we agree on removing `return None`, then that config-option is not necessary anymore, right? I mean, if the function does not have a Graph object to return (substitution by an empty Graph is not good either) and it should not return `None`, then what else can we do in the end except raise an exception (whether it be our own or bubbling up the original), right?

Hope this gets some things out of the way already. I will still get back on several other issues we were discussing (error handling, logging, printing, flags, etc.), either here or in our parallel discussion in issue #171.

ankostis: Thank you for the resources.
- The last one is outdated and fested with the ill-advice that troubles the newbies: if you have the code you are calling raise an exception, you should NOT catch it, unless you know what to do with it.  Besides, it's PY2.
- The 2nd from the end is a very nice discussion of a simple indexing operation: to raise or not (the inverse of the prebious).  For me, offering both possibilities, like `dict` does is the optimum.  That's simple.  But you do not ave this problem.
- The SO article confirms the pythonic consensus that state-return values are ugly.
- For the video , i dont have time, but i hope it goes further deep into the problem, and discussrs Error-Bariers (against the proliferation of unknown exception classes) or other architectural concerns and laws.

ankostis: > f we agree on removing return None, then that config-option is not necessary anymore, right? 

The suggested config-option was to retain BW-compatibilty.  If it's not needed, the it's totally fine (personally i'm ok with braking broken designs ;-)

peternowee: About that config-option `pydot.raise_errors`: We could still use it to create a transition period, with its default switching from `False` to `True` later than described above (e.g. 2.x.x). It would be similar (but inverse) to the debug/verbose flag discussed for printing in https://github.com/pydot/pydot/issues/171#issuecomment-612582791. For example, this other project started a transition period like that: https://github.com/sendgrid/sendgrid-python/issues/57.

peternowee: Sorry, I did not see your two comments before I sent out my last comment and went to sleep last night.

I will think about the BW-compatibility/transition period some more. Once we know where we want to go exactly, we might also have a better idea on how to get there. Also, I still hope to hear the opinion of the project maintainer, because he might know more about how pydot is used in practice.

> For the video , i dont have time, but i hope it goes further deep into the problem, and discussrs Error-Bariers (against the proliferation of unknown exception classes) or other architectural concerns and laws.

Not very deep, it was just an intro course. I could not find many results on Error-Bariers (or Error Barriers), do you perhaps mean [Fault Barriers](https://www.infoq.com/news/2007/01/effective-exceptions/)?

peternowee: Some more links for fault barriers:
- https://www.oracle.com/technical-resources/articles/enterprise-architecture/effective-exceptions-part1.html
- http://wiki.juneday.se/mediawiki/index.php/Design_patterns_-_Exceptions_patterns

ankostis: > Not very deep, it was just an intro course. I could not find many results on Error-Bariers (or Error Barriers), do you perhaps mean [Fault Barriers](https://www.infoq.com/news/2007/01/effective-exceptions/)?

Bullseye :+1:

dot_parser.py
peternowee: I just created a separate PR #227 with just the minimal bug fix for `TypeError` for a bug fix/point release.

peternowee:
I just created a separate PR #227 with just the minimal bug fix for `TypeError` for a bug fix/point release.

This PR (#219) will now target a future major/minor release and assume support for Python <3.5 is dropped. I plan to force-push new changes over this branch later, so until further notice, please do not merge.

dot_parser.py
peternowee: [Above](https://github.com/pydot/pydot/pull/219#discussion_r408918580), I wrote:
> I will split off the bug fix for the `TypeError` to a separate PR (as soon as we agree on the solution for that)

@ankostis: Though we did not agree on the details of it yet, I created PR #227 already, because the code of the bug fix there is no longer automatically part of the more extensive changes we are discussing here. I still welcome your comments there.

dot_parser.py
peternowee: Ok, following our discussion in https://github.com/pydot/pydot/issues/171#issuecomment-622489506, I will use logging instead.

dot_parser.py
peternowee: Here as well, you are right. Following our discussion in https://github.com/pydot/pydot/issues/171#issuecomment-622489506, I will use logging instead.

dot_parser.py
peternowee: Following our discussion in https://github.com/pydot/pydot/issues/171#issuecomment-622489506, I will use logging instead. I will use `%`-string formatting with the variables [passed as arguments](https://docs.python.org/3.8/howto/logging.html#logging-variable-data) to the logger to [defer actual formatting](https://stackoverflow.com/questions/5082452/string-formatting-vs-format/5082809#5082809).

peternowee:
Following the discussion in pydot/pydot#171, I have been working on better solutions for PRs pydot/pydot#218 and pydot/pydot#219. These solutions depend on dropping Python 2 support and having a logging facility. For this, I created PRs pydot/pydot#229 and pydot/pydot#231. Also, pydot/pydot#211 needs to be merged to make the Travis CI test results meaningful again.

I will now suspend my work until at least those 3 PRs (pydot/pydot#211, pydot/pydot#229, pydot/pydot#231) have been merged into the master branch. I remain available to rebase my PRs or advise about the order in which these and other PRs can best be merged.

Issue #187, opened 2018-10-18
quote_if_necessary is not bulletproof


Lucas-C:
Bug reproduction:
```
import pydot
g = pydot.Dot(root='"";""')
print(g.to_string())
g.write_png('out.png', prog='twopi')
```
Output:
```
digraph G {
root="";"";
}

['twopi', '-Tpng', '/tmp/tmp8wygs5cn'] return code: -11

stdout, stderr:
 b''
b''

Traceback (most recent call last):
  File "repro.py", line 4, in <module>
    g.write_png('out.png', prog='twopi')
  File ".../lib/python3.6/site-packages/pydot.py", line 1673, in new_method
    encoding=encoding)
  File ".../lib/python3.6/site-packages/pydot.py", line 1756, in write
    s = self.create(prog, format, encoding=encoding)
  File ".../lib/python3.6/site-packages/pydot.py", line 1884, in create
    assert p.returncode == 0, p.returncode
AssertionError: -11
```

That's because `quote_if_necessary('"";""')` returns `False` whereas some escaping needs to be done

ankostis:
It even fails `pydot.quote_if_necessary("a,b")` which is left unquoted.
We need a new goithub-label to mark all bugs with quoting,
and rewrite the `needs_quotes()`.

PaulSzymanski:
This bit me as well. Strings containing a colon (`:`) are not quoted but should be.

stevenjswanson:
I'll pile on as well.  The string `'\\"'` causes `to_string()` to generate invalid output.

```
import pydot

g = pydot.Dot()
g.add_node(pydot.Node("b", label='\\"'))
g2 = pydot.graph_from_dot_data(g.to_string())
assert g2 is not None, "Graph did not parse."
```

lkk7:
@stevenjswanson thanks for refreshing this issue! PyDot has been in a dead state for some time, but it was recently revived.
I'll look into this issue in free time.

ferdnyc:
I believe the situation is much improved in pydot 3.0.1, partly by the splitting of  ID and attribute quoting, but it seems there are still some improvements that could be made around escaping.

In graphviz, if you want quotation marks in a label you can either escape them, or use an HTML-like string. Either of these will correctly label a node `";"`:

```gv
node1 [label="\";\""];
node2 [label=<";">];
```

[...]

Issue #85, opened 2015-04-13
Multiple styles not correctly rendered (quotes missing)


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. add a node constructed like pydot.Node('A', shape='box', 
style='rounded,dotted')
2. use graph.write_raw() to emit the dot equivalent

What is the expected output? What do you see instead?
expected output:
  A [style="rounded,dotted", shape=box];

actual output:
  A [style=rounded,dotted, shape=box];

Note the lack of quoting. dot sees only "rounded" and ignores "dotted".

What version of the product are you using? On what operating system?
svn r29 from http://pydot.googlecode.com/svn/trunk

Please provide any additional information below.

Here is my complete test script:
"""
import pydot

graph = pydot.Dot(graph_type='digraph')
graph.add_node(pydot.Node('A', shape='box', style='rounded,dotted'))

graph.write_raw('twostyles.dot')
"""

and here is the complete dot file that I *expect* from that script:

"""
digraph G {
A [style="rounded,dotted", shape=box];
}
"""

```

Original issue reported on code.google.com by `gerg.w...@gmail.com` on 20 Sep 2013 at 6:02

PR #228, opened 2020-04-20
enh: support list styles in node/edge constructors

ankostis:
No test-case :-(

### DIFF
--- a/pydot.py
+++ b/pydot.py
@@ -593,6 +593,12 @@ class Node(Common):
 
             self.obj_dict = dict()
 
+            # It's only `style` attribute that maybe multi-valued.
+            #
+            style = attrs.get('style' )
+            if style and isinstance(style, (tuple, list)):
+                attrs[ "style" ] = ','.join(style)
+
             # Copy the attributes
             #
             self.obj_dict[ 'attributes' ] = dict( attrs )
@@ -726,7 +732,14 @@ class Edge(Common):
                   quote_if_necessary(dst))
         self.obj_dict['points'] = points
         if obj_dict is None:
+            # It's only `style` attribute that maybe multi-valued.
+            #
+            style = attrs.get('style' )
+            if style and isinstance(style, (tuple, list)):
+                attrs[ "style" ] = ','.join(style)
+
             # Copy the attributes
+            #
             self.obj_dict[ 'attributes' ] = dict( attrs )
             self.obj_dict[ 'type' ] = 'edge'
             self.obj_dict[ 'parent_graph' ] = None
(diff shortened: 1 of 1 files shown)


peternowee:
Hi @ankostis, thanks for this PR and sorry for my late reply. Hope you are doing ok.

I like your idea, but I will need some more time to think about it. It won't go in pydot 1.4.2 anyway, because that will just be a bugfix release.

With regard to pydot 2.0.0, I am hesitant if we should implement this already for only this single Graphviz-known DOT attribute (`style`), while there are [so many other attributes](https://www.graphviz.org/doc/info/attrs.html) for which there are also obvious conversions from certain Python types. If we do `style`, we should also be prepared to do other obvious ones. Of course, we don't have to do them all in one time, but still it is a goal that we would set then. Also, later changes to DOT attributes should also accept the same input (e.g. `set_style(...)`, `set('style', ...)`). On the other hand, we should also consider that not every DOT language user may be targeting Graphviz (e.g. use the `style` attribute for a custom purpose), so we might need an option to prevent these kind of conversions. I don't know if you would be interested in working more on all this? Because I feel I already have more than enough on my plate for pydot 2.0.0. Perhaps we could create a separate feature branch for this and merge it once it has a bit more substance, whether that be pydot 2.0.0 or a later version. Please let me know what you think.

Anyway, I will try to respond in more detail during development of pydot 2.0.0.

In case you got prompted to write this PR by the lack of quoting of comma-separated strings: I certainly intend to get that solved with pydot 2.0.0. I see you already made a comment about it in issue https://github.com/pydot/pydot/issues/187#issuecomment-623127490. Issue https://github.com/pydot/pydot/issues/85 is more specific. The problem is with the regex that checks if the string should get quotes or not.

About the second commit you made (commit c8ff1330): I think that can be applied independent of this PR, right? Perhaps you can open a separate PR for it and that can go in pydot 2.0.0 already. Just two remarks:
- The commit only changes `Edge.__init__()`, but it seems the same change could be made in `Node.__init__()` and `Graph.__init__()` as well.
- Maybe improve the description a bit, because the current one ("fix: copying **kw now really needed") is a bit cryptic. Perhaps something like "Drop excess copying of fresh keyword dictionaries".

Thanks, take care,
Peter

ferdnyc:
Oh, I think this is now fixed — if I'm reading this right, it's the same issue that my #320 addressed.

Ironically enough, it was the _numeric_ value handling that was tripping it up — the regular expressions used by `quote_if_necessary`, seemingly written for European-style number formatting, were quoting values containing `.` but _not_ quoting values containing `,`. Which bled into its handling of `style` attribute quoting, and obviously caused problems because attributes are comma-separated in graphviz.

But, like I said, I believe now fixed by #320.

Issue #205, opened 2019-01-28
Added Graphviz too Path but pydot cannot find dot excuable correctly


QingGo:
I use anaconda python on windows 10 platform, install pydot by

`
pip install pydot
`
then install Graphviz by install package download from https://www.graphviz.org/download/.
and then add the install location to Path:
`
E:\Program Files (x86)\Graphviz2.38\bin
`
but pydot cannot find dot excuable correctly because of the code in pydot below as I think:

``` python
def get_executable_extension():
    # type: () -> str
    if is_windows():
        return '.bat' if is_anacoda() else '.exe'
    else:
        return ''

```
There are not excuable file name dot.bat, only dot.exe in my path. Even I use anaconda, but I am not install graphviz by conda.
And the error info is confusing：
`
FileNotFoundError: [WinError 2] "dot" not found in path.
`
Pydot just can not find **dot.bat** but not **dot**.

QingGo:
Finally, I use
```
conda install -c anaconda graphviz
```
to install graphviz and then everything going well.

prmtl:
Yes, pydot installed via conda it supports graphviz installed by conda only.

I agree that the error message should take different installation methods into account and be improved.

QingGo:
Thank to answer. 
Actually I install pydot via pip but not conda. Even I use the python distribution of Anaconda but I merely use conda as package manager. 
I have solve this problem, but I just want to point out this ploblem here to save others time which face the same problem.

Ashafix:
Just ran into the same problem, installation via `pip` but using it via an Anaconda notebook. `dot.exe` is present but it searched for `dot.bat`.
I am not sure if `get_executable_extension` is needed at all. If `dot.exe` or `dot.bat` is in `PATH` it doesn't matter whether `subprocess.Popen` is with or without extension.

peternowee:
Copy of my comments in duplicate issue pydot/pydot#210 _(bugs in windows and anaconda env)_:

> I don't have the solution, but just wanted to share some information related to this issue that I came across while investigating another issue:
> 
> - `pydot`'s use of the `.bat` extension for Graphviz on Anaconda on Windows was mostly discussed in [pydot/pydot-ng#52](https://github.com/pydot/pydot-ng/pull/52) by @prmtl and the ever eloquent @leycec. It mentions some reasons for why "Python users on Windows typically leverage `conda` rather than `pip`", so check it out.
[...]

PR #236, opened 2020-09-01
fix graphviz's suffix problem

Clinale:
modify the function _**is_anaconda**_

### DIFF
--- a/pydot.py
+++ b/pydot.py
@@ -97,7 +97,10 @@ def is_windows():
 
 def is_anaconda():
     # type: () -> bool
-    return os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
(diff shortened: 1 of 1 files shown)


peternowee:
@Clinale The fix you proposed https://github.com/pydot/pydot/issues/205#issuecomment-645734478, thank you for turning it into a PR so quickly. Just some formatting issues, could you please address these:
- Remove (instead of comment out) the old line and the empty line in between (remove the current new lines 100 and 101.)
- Remove the trailing space after `import glob`.

You can overwrite (force-push) over this branch with an improved commit. While you are at it, maybe you can add "Fixes pydot/pydot#205" in the body of the commit message.

peternowee:
Superseded by #237.

PR #237, opened 2020-09-03
fix graphviz's suffix problem / Fixes #205

### LINKED ISSUE #205: Added Graphviz too Path but pydot cannot find dot excuable correctly

Clinale:
fixes #205

### DIFF
--- a/pydot.py
+++ b/pydot.py
@@ -97,7 +97,8 @@ def is_windows():
 
 def is_anaconda():
     # type: () -> bool
-    return os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
+    import glob
+    return glob.glob(os.path.join(sys.prefix, 'conda-meta\\graphviz*.json')) != []
 
 
 def get_executable_extension():


peternowee:
@Clinale Great work, thank you! Let's hope this solves the problem, at least for most people. Thanks!

PR #231, opened 2020-05-15
Implement basic library logging

peternowee:
_From the commit message:_

pydot currently prints some error messages to standard output (`stdout`). Printing anything is considered bad practice for libraries ([SO][1], [reddit][2]). This becomes even more disturbing if the prints contain long Graphviz output or DOT strings.

[1]: https://stackoverflow.com/questions/4201856/error-handling-strategies-in-a-shared-library-c
[2]: https://www.reddit.com/r/C_Programming/comments/97ebfj/redirect_stdout_and_stderr_to_a_buffer/e47n9kh/

This commit lays the groundwork for switching from printing to logging by providing a basic library logging setup and some user documentation on how to read the logs. Later commits will replace the actual printing by logging.

Discussion: https://github.com/pydot/pydot/issues/171.

_Additional remarks:_

- This PR depends on pydot/pydot#229, because the test uses [importlib.reload](https://docs.python.org/3/library/importlib.html#importlib.reload) and [unittest.assertLogs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs), which were both introduced in Python 3.4.
- When merged with PRs pydot/pydot#211 and pydot/pydot#229, all tests pass on my laptop (Debian buster 10.4, Linux 4.19.118-2 amd64, Python 3.7.3, Graphviz 2.40.1 (20161225.0304), pyparsing 2.4.7).
- On my Travis CI account, when merged with pydot/pydot#211 and pydot/pydot#229, all tests pass. See https://travis-ci.com/github/peternowee/pydot/builds/166387174.
- On my AppVeyor account, when merged with pydot/pydot#211 and pydot/pydot#229, all tests pass except the ones mentioned in known issue pydot/pydot#213. See https://ci.appveyor.com/project/PeterNowee/pydot/builds/32905623/job/67uhk54284opixkn.
- Let me know if you would like to have the documentation in a separate file, such as TROUBLESHOOTING.md. I am working on other PRs that will cover troubleshooting Graphviz errors and DOT string errors as well, so maybe it is a good idea to split it off. I also noticed your earlier proposal to switch to Sphinx documentation in pydot/pydot#161. If you want to give that a second chance, let me know and I can try to adapt to that format as well.
- Let me know if you start merging PRs and want me to rebase this branch or if you need help determining in which order to merge all the open PRs.
- Reviews welcome!

CC: @ankostis

Issue #230, opened 2020-05-11
Restructure package and module organization


peternowee:
This issue is split off from a discussion in pydot/pydot#171 (and pydot/pydot#218 and pydot/pydot#219) in which @ankostis and I were discussing error handling and logging. The question was raised if  `pydot.py` and `dot_parser.py` should share a common custom exceptions hierarchy and a common parent logger. This in turn raised questions about the current package structure.

@peternowee wrote on [2020-05-01](https://github.com/pydot/pydot/issues/171#issuecomment-622489506):
> > I run into even bigger questions, like: Why are pydot.py (from where Graphviz is called) and dot_parser.py (from where PyParsing is called) separate? Does sharing an exception or exception hierarchy between the two fit in with that? Should we define the exception hierarchy in a separate file? What to do with the existing exceptions, which have some problems already? Naming issues, etc. etc.

@ankostis replied, also on [2020-05-01](https://github.com/pydot/pydot/issues/171#issuecomment-622531736):
> ### Package & module organization
> 
> I agree, the current organization of python files is "unusual", and complicates crafting a sensible exception hierarchy.
> The majority of distributions in PyPi are either a single python-file (a module),
> or a package (a folder with `__init__.py` **package-module**) containing more modules files inside.
> 
> If `dot_parser.py` us considered a "private" file, it' i's easy to retrofit the existing 2 modules into a more conventional format like that:
> 
> ```
> pydot/
>   +--__init__.py  # all "public" objects must be defined/imported in this module
>   +--dot_parser.py 
>   +--base.py      # read below 
> ```
> 
> The Exceptions should be definitely defined in the package-module.
> 
> It is tempting to copy-paste the whole `pydot.py` contents on the package-module
> but i would recommend against it.
> 
> > TIP1:
> > When a user imports a module further into a package, the package-module gets import
> > in its entirety - so if it is "heavy" it delays the import.  That is important when
> > the different functionality is split in multiple submodules of the package.
> > Example:
> > A package contains 3 sub-modules: base.py, sphinx_extension.py
> > If `base.py` contains also code for irrelevant http-server communications,
> > whenever `from <package> import sphinx_extension` code is trun,
> > it has to pay the price for this never-to-be-used http-server functionality.
> 
> > TIP2
> > The modules are split based on the 3rd-party imports.  If a module is completely disparate,
> > it might become an "extras" in the `setup.py`.
> 
> Copy all in the `base.py` (or choose another name for this module, just my convention, try `core.py`),
[...]

PR #229, opened 2020-05-11
Drop Python 2 and 3.4 support

peternowee:
**UPDATE 2021-06-25**: Updated description in my <a href="https://github.com/pydot/pydot/pull/229#issuecomment-868431830">comment of 2021-06-25</a> below.

<details><summary>Original description (outdated)</summary>

_1st commit:_ **Drop Python 2 and <=3.4 support, add 3.8 support**

Why support for Python 3.5 is not dropped yet:

- We do not urgently need Python 3.6+ specific functionality.
- Python 3.5 is still receiving security fixes until 2020-09-13.[[1]] [[2]]
- Recent PyPI download [statistics for pydot][3] show that 20 to 40% of the users are still on Python 3.5.

For further details, see [pydot/pydot#171][4].

_2nd commit:_ **Drop Travis CI tests Python 2, 3.4, add Python 3.8**

Also removes the workaround for the Python 3.7 test environment added by commit eb7aef13. It is no longer needed by travis-ci.com.

_3rd commit:_ **Change AppVeyor tests to Python 3.5 and 3.8**

Drop the test for Python 2 now we no longer support that.

Also:
- Explicitely choose the image to be used (Visual Studio 2019), as the default image (Visual Studio 2015) does not have [Python 3.8 *in Miniconda*][5].
- Add [`fast_finish` option][6] to prevent running subsequent jobs if the first job fails.
- Only test the oldest and the newest supported Python versions (currently Python 3.5 and 3.8). Intermediate versions are tested in Travis CI already.
- Correct a typo in the `init` that prevented correct printing of environment variables.

_Additional remarks:_

- This PR targets a future major release of pydot.
- Earlier discussion was in issue pydot/pydot#171.
- No code changes yet. This is just to declare that Python 2 compatibility will break soon, because of other changes that I am working on. I do not have time for a cleanup of Python 2-supporting code right now and I also do not think that is necessary for this to merge.
- All tests pass on my laptop (Debian buster 10.4, Linux 4.19.118-2 amd64, Python 3.7.3, Graphviz 2.40.1 (20161225.0304), pyparsing 2.4.7).
- Travis CI runs all the test environments 3.5 through 3.8 and all of them succeed when run together with pydot/pydot#211.
- AppVeyor starts the tests for 3.5, and also for 3.8 when `fast_finish` is off, but both still fail because of issue pydot/pydot#213.
- Let me know if you start merging PRs and want me to rebase this branch or if you need help determining in which order to merge all the open PRs.
- Reviews welcome!

[...]

Issue #222, opened 2020-01-10
pydot.Dot.create(pydot.Dot()) returns assertion error


pj1989:
The command
```
pydot.Dot.create(pydot.Dot())
```
returns 
```
"dot" with args ['-Tps', 'C:\\Users\\LUCA~1.AME\\AppData\\Local\\Temp\\tmp9jsxm52p'] returned code: 1

stdout, stderr:
 b''
b''

Traceback (most recent call last):

  File "<ipython-input-11-2d049bb9ae08>", line 1, in <module>
    pydot.Dot.create(pydot.Dot())

  File "C:\Users\Luca.Amerio\AppData\Local\Continuum\anaconda3\lib\site-packages\pydot.py", line 1945, in create
    assert process.returncode == 0, process.returncode

AssertionError: 1
```

Running `pydot` version 1.4.1

pj1989:
The problem can be solved commenting line 117 
```
program += extension
```
in `pydot.py`.

In other words, the subprocess call does not need to include the program extension

peternowee:
Judging by the discussion in the keras-issue you link to, which mentions `dot.bat` and `dot.exe`, and by the nature of your fix, I think you ran into issue pydot/pydot#205. I will mention your fix there. If you agree, you can close this issue and continue discussion in pydot/pydot#205.

peternowee:
As mentioned in my previous comment, I am closing this as a suspected duplicate of pydot/pydot#205.

PR #239, opened 2020-09-24
AssertionError -> RuntimeError

peterjc:
Currently getting a macOS failure within a larger software stack, and do not see the message printed to stdout.

Replace a traceback with "AssertionError: 1" with the meaningful message, since this is not useful:

```
...
  File "/Users/xxx/miniconda3/lib/python3.7/site-packages/networkx/drawing/nx_pydot.py", line 257, in graphviz_layout
    return pydot_layout(G=G, prog=prog, root=root)
  File "/Users/xxx/miniconda3/lib/python3.7/site-packages/networkx/drawing/nx_pydot.py", line 306, in pydot_layout
    D_bytes = P.create_dot(prog=prog)
  File "/Users/xxx/miniconda3/lib/python3.7/site-packages/pydot.py", line 1723, in new_method
    format=f, prog=prog, encoding=encoding)
  File "/Users/xx/miniconda3/lib/python3.7/site-packages/pydot.py", line 1945, in create
    assert process.returncode == 0, process.returncode
AssertionError: 1
```

### DIFF
--- a/pydot.py
+++ b/pydot.py
@@ -1943,6 +1943,12 @@ class Dot(Graph):
             )
             print(message)
 
-        assert process.returncode == 0, process.returncode
+        assert process.returncode == 0, (
+                '"{prog}" with args {arguments} returned code: {code}'.format(
+                    prog=prog,
+                    arguments=arguments,
+                    code=process.returncode,
+                )
+            )
 
         return stdout_data


peterjc:
With the change I get a useful exception message:

```
RuntimeError: "fdp" with args ['-Tdot', '/var/folders/xq/zhs3hy7x5cv50svc4s7q2_9h0000gn/T/tmp0y9mct4x'] returned code: 1
```

Reproducible outside of pydot,

```
$ fdp -Tdot /var/folders/xq/zhs3hy7x5cv50svc4s7q2_9h0000gn/T/tmp0y9mct4x
Warning: Could not load "/Users/xxx/miniconda3/lib/graphviz/libgvplugin_pango.6.dylib" - It was found, so perhaps one of its dependents was not.  Try ldd.
Warning: Could not load "/Users/xxx/miniconda3/lib/graphviz/libgvplugin_neato_layout.6.dylib" - It was found, so perhaps one of its dependents was not.  Try ldd.
There is no layout engine support for "fdp"
Use one of: circo dot fdp neato nop nop1 nop2 osage patchwork sfdp twopi
```

This looks like a condaforge packaging issue.

peternowee:
Hi @peterjc, thank you for your PR. The limited value of the `AssertionError` message after Graphviz errors was earlier discussed in pydot/pydot#203 and pydot/pydot#171. I provided some instructions on how to investigate it in https://github.com/pydot/pydot/issues/203#issuecomment-533164709 and I am preparing a better solution in PR pydot/pydot#218, which will involve logging the error details (more details than are currently printed even) and raising an exception with a clearer error message, as was explained to me by @ankostis in pydot/pydot#171.

Why pydot/pydot#218 is currently closed: I closed it last May 2020, because groundwork that it depended on (pydot/pydot#229, pydot/pydot#231) was not getting merged. However, at the end of August 2020, I was finally granted commit rights, so things got moving again, but I need some time now to reopen pydot/pydot#218 again, because I still need to adapt it to the outcomes of the discussion in pydot/pydot#171.

With regard to the time frame: My main focus right now is on getting acquainted with all the open issues and PRs, sorting out which ones are most urgent and can be fixed without the risk of new regressions and then getting a bugfix release 1.4.2 out. After that, I was planning to branch off the 1.4.x series into a separate git branch (the last to support Python 2), and open up the master branch for merging of all kinds of larger changes such as logging and error reporting. This should all happen within the next few weeks, if all goes well. Note, however, that it will take many months for those large changes to become part of an official release (2.0.0 or 1.5.0), because some other large changes, such as improving DOT syntax compliance, API and object model changes for example, will need that more discussion, development and testing time.

The question then becomes if changes to the error reporting should be made in pydot 1.4.2 already. I especially have doubts about changing the exception class, but also about moving the entire `message` string from printing to the exception message, as you propose. Both seem quite big changes for a mere point release.

Some side notes:
- The Graphviz stdout in `message` may still contain a long DOT string or binary data, even on errors. See the table at the bottom of https://github.com/pydot/pydot/issues/215#issuecomment-536198725.
- pydot/pydot#203 showed that in many cases not Graphviz stdout, but the pydot DOT string was needed to find the root cause. In the end, I gave up on my original proposal for pydot/pydot#218 (https://github.com/pydot/pydot/commit/51d5dcdc25c1dc5fdc4aceaa2f5f69099a2c3030) to make the pydot DOT string part of `message` as well.
- In the upcoming new proposal for pydot/pydot#218, for 2.0.0 or 1.5.0, details including the pydot DOT string and Graphviz invocation details, error code, stdout, stderr will get logged at DEBUG level. The exception message will just give a summary (possibly the Graphviz error code and stderr) and point to the log and the (also new) troubleshooting documentation for finding the rest. The exception class will change from `AssertionError` as well then.

Maybe a more limited change could be in 1.4.2 already, like I did with parsing errors (large change pydot/pydot#219 for 2.0.0/1.5.0 and limited change pydot/pydot#227 for 1.4.2), but I am wondering what that limited change could then be in this case. I would prefer, for 1.4.x, to at least keep the exception class the same (`AssertionError`), to not introduce logging and to not drop the printing. Maybe we can extend the exception message, but I have doubts about putting all of Graphviz stdout in there. Maybe add only Graphviz stderr? Assuming Graphviz (`fdp`) printed its missing library errors to stderr, I guess that could have helped in your case. But not in the cases in pydot/pydot#203 that needed the pydot DOT string, or in case of missing executable as in pydot/pydot#222 for example. And in the end, you still found the root cause for your case as well, so is it worth all the trouble? I will think about it some more, but maybe you can also give your opinion, and/or maybe even change your PR already (you can just force-push over it).

Good to see you already found the underlying cause, btw. Hope it gets solved soon.

peterjc:
I didn't search hard enough - I agree, this PR is a quick and dirty solution, and you are trying to solve a bigger problem.

I think if I restore the print, and shorten the exception to just ``"{prog}" with args {arguments} returned code: {code}"`` that would be a good short term change worth considering for a point release.

peternowee:
Yes, in an `AssertionError` as before, right? No stderr, I see, and indeed it seems better to leave it out, because it can also span many lines, see https://github.com/pydot/pydot/issues/203#issuecomment-533762637 for example.

Thanks!

peterjc:
I think ``AssertionError`` is the wrong choice - people do run Python with asserts disabled. But if you prefer, it can stay as that.

ankostis:
Just a quick note, sure `assert` keyword is wrong for all "production" purposes, but here it is raising explicetely an `AssertionError`, correct?  Even with disabled assertions, it would still work.

Anyhow, this exception class should change, it was ill-conceived in the first place, 
but since that would constitute a bw-incompatible change, better wait for a major version bump.

peternowee:
First of all, I think we all agree that `AssertionError` and the whole `assert` were wrong to begin with. This started with commits 9b3c1a1 and bc639e7 of 2016 (v1.2.0). My plan is to replace it in pydot 2.0.0/1.5.0 with raising a custom exception, as discussed in pydot/pydot#218 and pydot/pydot#171. We still need to see what the exception hierarchy will look like exactly, though. That touches on the issue of the relationship between `pydot.py` and `dot_parser.py` (pydot/pydot#230), so may take some time to refine. Also I was hoping to do some PY2 code clean up, coding style corrections, etc., all while trying to adopt the best practices of forks like pydot-ng, pydotplus and pydot's dev branch. I was planning to work out the details of all that after the release of 1.4.2, probably by posting PRs for discussion in weeks/months thereafter. I intend to keep using PRs (not push to master directly) for these kinds of major changes and wait at least one week to see if they lead to any discussion before merging them. I do not think I will keep mentioning people when I submit my PRs, because these mentions often go unanswered, so I am not sure if that is appreciated. But if you 'watch' the Github.com repository, you should get notified automatically. Of course, feel free to make proposals and kick off the discussion earlier, but I am not sure how much I can participate.

Then, about changing the error reporting for the upcoming point release 1.4.2, I would just like to keep the change as minimal as possible. Even changing from `assert` to `raise AssertionError` will change the behavior for users who normally run with assertions disabled (e.g. [those running with optimization](https://docs.python.org/3.8/reference/simple_stmts.html#the-assert-statement)), as you both noted. They will start getting errors whereas before they may still have received output from Graphviz (see example 3 in https://github.com/pydot/pydot/issues/215#issue-499470442). The workflow of such users would break because of this change, so I have doubts about doing that in a point release. Of course, [every change break someone's workflow](https://xkcd.com/1172/), so we have to make that choice sometimes, but it is just difficult since we have limited knowledge of how others use pydot. Opinions (or further reduction of this PR) welcome.

peterjc:
That makes sense to me. Even smaller change coming up - just replacing the assert message.

peternowee:
Ok, thanks for understanding. Code looks good to me now, though now the thing is I prefer to have the 3 commits squashed into one and improve the commit message a bit. This is the first project I maintain, so I do not have much experience with doing that on Github yet.

Github shows me a nice "Squash and merge" button, but there are [reports that it is unreliable](https://github.com/isaacs/github/issues/1368) when it comes to the authorship fields Author and Commit. Though it is common practice that the Commit-field changes to person making the commit, the "Squash and merge" button is said to sometimes even touch the Author-field, which is of course not my intention.

Alternatively, I can squash and commit locally and then push to Github, but then I think this PR will not be considered 'Merged' and need to be 'Closed', all in Github terms. See https://github.com/pydot/pydot/pull/209#issuecomment-682214089 for how that bit me the first time I first locally merged (rebased in that case, actually) someone else's PR.

Also, you might want to have a say in what the commit message will be.

If I would do it, I would make it something like this:

```
Improve traceback error message on Graphviz errors

For reporting Graphviz errors, improve the "AssertionError" message
in the traceback by including program name and arguments. This is the
same as the first line of the error message that is printed to stdout,
but not followed by the (potentially long) Graphviz stdout and stderr
output.

Related issue: https://github.com/pydot/pydot/issues/203
PR: https://github.com/pydot/pydot/pull/239
```

You seem to have a lot more experience on with this, so hope you give me some advice and let me know what you prefer.

peterjc:
I am aware of complications when the author has deliberately used another email address in the git commits - rare but very frustrating when it happens. GitHub's squash-and-merge should be safe here (I use it frequently), especially as I've used the same details as my GitHub account. When the admin does a squash-and-merge, they get to (re)write the commit message within GitHub - and your wording sounds great.

peterjc:
Oh, and I've found only fast forward local merges work nicely with GitHub spotting and marking the associated PR as merged.

peternowee:
Yes, finally. Thanks a lot for your patience and help! And for the fix, of course, hope it will help people in troubleshooting!

peterjc:
No problem - I've used pydot on and off for years, happy to see it being actively maintained again. Thank you!

PR #241, opened 2020-10-01
bug fixes, API changes, documentation

johnyf:
- API:
  - require two positional arguments `src` and `dst` in `get_edge`
  - catch `ImportError`, instead of `Exception`
- BUG:
  - return both aligned and reversed edges from `get_edge`
  - allow semicolon as attribute separator
- DOC:
  - improve warning message about `dot_parser`
  - clarify the effect of calling the method `set_node_defaults`
- REL:
  - update file `LICENSE`

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
-Copyright (c) 2014 Carlos Jenkins
-Copyright (c) 2014 Lance Helper
-Copyright (c) 2004 Ero Carrera
+Copyright (c) 2004 by Ero Carrera
+Copyright (c) 2014 by Lance Helper
+Copyright (c) 2014 by Carlos Jenkins
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2004 by Ero Carrera
+Copyright (c) 2004-2020 by Ero Carrera
 Copyright (c) 2014 by Lance Helper
 Copyright (c) 2014 by Carlos Jenkins
 
--- a/dot_parser.py
+++ b/dot_parser.py
@@ -471,7 +471,7 @@ def graph_definition():
 
         node_id = (ID + Optional(port))
         a_list = OneOrMore(ID + Optional(equals + righthand_id) +
-            Optional(comma.suppress())).setName("a_list")
+            Optional((semi | comma).suppress())).setName("a_list")
 
         attr_list = OneOrMore(lbrack.suppress() + Optional(a_list) +
             rbrack.suppress()).setName("attr_list")
--- a/pydot.py
+++ b/pydot.py
@@ -13,7 +13,7 @@ import warnings
 
 try:
     import dot_parser
-except Exception as e:
+except ImportError as e:
     warnings.warn(
         "Couldn't import dot_parser, "
         "loading of dot files will not be possible.")
--- a/pydot.py
+++ b/pydot.py
@@ -1309,25 +1309,15 @@ class Graph(Common):
 
         return False
 
-
-    def get_edge(self, src_or_list, dst=None):
-        """Retrieved an edge from the graph.
-
-        Given an edge's source and destination the corresponding
-        Edge instance(s) will be returned.
+    def get_edge(self, src, dst):
+        """Return edges with from node `src` to node `dst`.
 
         If one or more edges exist with that source and destination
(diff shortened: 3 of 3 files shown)


peternowee:
Hi @johnyf, thanks for rebasing the `dev`-branch. These are the ones I would like to merge for the upcoming 1.4.2 release, with some remarks:

- DOC: clarify the effect of calling `set_node_defaults`

  (c9a002b in this PR pydot/pydot#241, 71df273 + 56ea713 in the [old `dev`](https://github.com/pydot/pydot/tree/dev) branch.)

  Remarks:
  
  - How about "nodes added to the graph after ..." instead of "nodes created after ..." and "only to" instead of "to only"?
  - What docstring line length are we using? I see them sometimes get wrapped at 50 or 60 characters already, which seems a bit short to me. How about 72 characters?
  
  Alternative:
  
  ```python
          """Define default node attributes.
  
          These attributes apply only to nodes added to the graph after
          calling this method.
          """
  ```

- MAI: improve warning message about `dot_parser`

  (9dea3fc in this PR pydot/pydot#241, 8f9f7c6 in the [old `dev`](https://github.com/pydot/pydot/tree/dev) branch.)

  Without the related commit "API: catch `ImportError` instead of `Exception`".

Could you rebase them and maybe split them off to one or two separate PRs? I will contact you about the other commits again later.

johnyf:
Thank you for reviewing the changes. I will rebase the commits mentioned above and open another pull request.

The choice "created" was to emphasize that the nodes are added for the first time after setting node defaults. About the "only to", this phrasing has also a different meaning according to the Wiktionary (item 5 in the list at https://en.wiktionary.org/wiki/only#Adverb).

I use short lines to allow opening several tabs of code side by side in an editor. A line width of 72 characters for docstrings agrees with PEP 8 (https://www.python.org/dev/peps/pep-0008/#maximum-line-length).

peternowee:
> Thank you for reviewing the changes. I will rebase the commits mentioned above and open another pull request.

Ok, thanks.

> The choice "created" was to emphasize that the nodes are added for the first time after setting node defaults. 

Ah ok, I see. But the problem is that "created" can also be understood as the nodes being instantiated and with that meaning the statement would not be true, because the node defaults will also apply to nodes instantiated before setting the defaults, but added to the graph after setting the defaults.

```python
import pydot
g=pydot.Dot()
n1 = pydot.Node('n1')
n2 = pydot.Node('n2')
g.add_node(n2)
g.set_node_defaults(color='blue')
g.add_node(n1)
print(g)
g.write_png('test.png')
```

```dot
digraph G {
n2;
node [color=blue];
n1;
}
```

test

> About the "only to", this phrasing has also a different meaning according to the Wiktionary (item 5 in the list at https://en.wiktionary.org/wiki/only#Adverb).

I do not see how "only to" could be understood to introduce such a construction here, because the rest of the sentence does not have an infinitive verb. See the examples in that item 5 ("to lose", "to betray") and on https://en.wiktionary.org/wiki/only_to ("to discover"). My feeling says "only to" would be better here, but if you feel the opposite I won't hold up merging this PR for it. The meaning is clear enough.

> I use short lines to allow opening several tabs of code side by side in an editor. A line width of 72 characters for docstrings agrees with PEP 8 (https://www.python.org/dev/peps/pep-0008/#maximum-line-length).

Yep, thanks.

peternowee:
Hi @johnyf, hope you are ok. Do you still want to provide new commits for the two commits discussed above?

I sent you an email on 2020-12-22, btw.

peternowee:
@johnyf: I would like to move on with pydot 1.4.2. Do you still want to rebase and change these two commits yourself? Otherwise, I propose I just rebase and merge them as you provided them and then add my further changes in subsequent commits myself. Please let me know before 2021-01-22, also if you need more time.

peternowee:
I cherry-picked the following two commits to master for inclusion in pydot 1.4.2:

- Commit "MAI: improve warning message about dot_parser" (9dea3fc in this PR #241, 8f9f7c6 in the [old `dev`](https://github.com/pydot/pydot/tree/dev) branch) was merged as pydot/pydot@22cce07.
- Commit "DOC: clarify the effect of calling `set_node_defaults`" (c9a002b in this PR #241, 71df273 + 56ea713 in the [old `dev`](https://github.com/pydot/pydot/tree/dev) branch) was merged as pydot/pydot@44ab099. I added my corrections in the subsequent commit pydot/pydot@79ae71c.

I will look at the rest later. Thank you and hope you are ok. Sorry for my nitpicking.

Issue #92, opened 2015-04-13
Simplify throws NameError, and, as implemented, doesn't work


GoogleCodeExporter:
What steps will reproduce the problem?
1. invoke `set_simplify(true)` on a graph
2. invoke method `to_string()`
3. profit

What is the expected output? What do you see instead?
Extraneous edges removed from dot output

What version of the product are you using? On what operating system?
1.0.2

Please provide any additional information below.

This is a patch which both resolves the `NameError` and enables the desired behavior:

```patch
--- /usr/lib/python2.6/site-packages/pydot.py   2014-06-12 06:22:57.000000000 
+0400
+++ virtenv/lib/python2.6/site-packages/pydot.py    2014-06-12 06:28:42.515864719 
+0400
@@ -1455,11 +1455,11 @@

                 edge = Edge(obj_dict=obj)

-                if self.obj_dict.get('simplify', False) and elm in edges_done:
+                if self.obj_dict.get('simplify', False) and 
edge.obj_dict['points'] in edges_done:
                     continue
-
+
                 graph.append( edge.to_string() + '\n' )
-                edges_done.add(edge)
+                edges_done.add(edge.obj_dict['points'])

             else:
```

Original issue reported on code.google.com by `TheSoup...@gmail.com` on 12 Jun 2014 at 2:30

Attachments:
- [pydot_simplify.patch](https://storage.googleapis.com/google-code-attachments/pydot/issue-92/comment-0/pydot_simplify.patch)

peternowee:
<span id="pn0000"></span>
### Table of contents

- [The bug](#pn0100)
- [Patch 1 by the OP](#pn0300)
- [Patch 2 by the OP](#pn0500)
- [Follow-up](#pn0700)

<span id="pn0100"></span>
### The bug [¶](#pn0100) [^](#pn0000)

#### Minimal reproducible example (MRE)

The following test case for the test suite contains a minimal reproducible example for the bug reported at the top of this post:

```python
    def test_graph_simplify(self):
        # Fail example: pydot 1.0.2. GH pydot/pydot#92 OP patch 1.
        g = pydot.Graph()
        g.add_edge(pydot.Edge('a', 'b'))
        g.add_edge(pydot.Edge('a', 'b'))
        g.add_edge(pydot.Edge('b', 'a'))
        g.add_edge(pydot.Edge('b', 'a'))
        test_combinations = [
            ('graph', False,
             'graph G { a -- b; a -- b; b -- a; b -- a; }'),
[...]

PR #247, opened 2020-11-27
Add test for simple/simplified graphs

peternowee:
This test tests the output of `Graph.to_string()` for graphs with multiple (parallel) edges and combinations of `simplify` set to `True` or `False` and the graph being directed or undirected.

This test was originally created for issue [pydot/pydot#92][1] ("Simplify throws NameError, and, as implemented, doesn't work"). It can also help catch regressions when doing future work on `Graph.to_string()`, `Edge.__eq__()` or `Edge.__hash__()`.

[1]: https://github.com/pydot/pydot/issues/92


peternowee:
This PR was merged as commit 5c9b2ce.

PR #249, opened 2020-11-27
Add tests for Edge equality and hashing

### LINKED ISSUE #92: Simplify throws NameError, and, as implemented, doesn't work

peternowee:
These tests test `Edge` equality comparison and hashing for regressions and point out some current issues.

For equality comparison (`Edge.__eq__()`):

- Our current definition of edge equality: Same endpoints and, if the parent graph is directed, the same order of endpoints. (Changes to that definition may require changing some of these tests as well.)
- [Identical objects should compare equal][1] (reflexivity).
- [Comparisons should be symmetric][1].
- Rich comparison methods should [return `NotImplemented` if they do not implement the operation for the operands provided][2] (see also [here][3]).
- Availability of equality comparison before the edge has a parent graph.

For hashing (`Edge.__hash__()`):

- [Hashable objects which compare equal must have the same hash value][4] (see also [here][5]).
- [The hash value of an hashable object should never change during its lifetime][4] (see also [here][5]).
- Try to prevent hash collisions, because they may degrade performance.
- Availability of the hash method before the edge has a parent graph.

Some tests are based on problems seen with the bug and patches reported in issue [pydot/pydot#92][6]. Some other tests are based on examples from articles by [Aaron Meurer][7] and [Roy Williams][8] and adapted to our `Edge` class.

The tests that are currently failing have [`expectedFailure`][9] decorators that will be removed as their underlying issues are addressed in future commits.

The PR for this commit is [pydot/pydot#249][10]. It should get links to spin-off issues and PRs that discuss the various issues brought up by these tests in more detail.

[1]: https://docs.python.org/3.9/reference/expressions.html#value-comparisons
[2]: https://docs.python.org/3.9/reference/datamodel.html#the-standard-type-hierarchy
[3]: https://docs.python.org/3.9/reference/datamodel.html#object.__eq__
[4]: https://docs.python.org/3.9/glossary.html#term-hashable
[5]: https://docs.python.org/3.9/reference/datamodel.html#object.__hash__
[6]: https://github.com/pydot/pydot/issues/92
[7]: https://www.asmeurer.com/blog/posts/what-happens-when-you-mess-with-hashing-in-python/
[8]: https://eng.lyft.com/hashing-and-equality-in-python-2ea8c738fb9d
[9]: https://docs.python.org/3.9/library/unittest.html#unittest.expectedFailure
[10]: https://github.com/pydot/pydot/pull/249

_Additional remark_: This PR is based on the fix of commit pydot/pydot@e4742e91 of PR pydot/pydot#248, because if it was based on current `master` without that fix, the results for `test_edge_equality_basics_3_same_points_not_not_equal` would differ between Python 2 and Python 3 and it would be difficult to mark it as an expected failure only for Python 2. Please post comments on commit pydot/pydot@e4742e91 in pydot/pydot#248, not here.

### DIFF
--- a/test/pydot_unittest.py
+++ b/test/pydot_unittest.py
@@ -7,6 +7,10 @@
 from __future__ import division
 from __future__ import print_function
 import argparse
+try:
+    from collections.abc import Hashable  # PY3
+except ImportError:
+    from collections import Hashable  # PY2
 from hashlib import sha256
 import io
 import os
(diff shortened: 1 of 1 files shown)


ferdnyc:
> For hashing (`Edge.__hash__()`):
> 
> * [Hashable objects which compare equal must have the same hash value](https://docs.python.org/3.9/glossary.html#term-hashable) (see also [here](https://docs.python.org/3.9/reference/datamodel.html#object.__hash__)).
> * [The hash value of an hashable object should never change during its lifetime](https://docs.python.org/3.9/glossary.html#term-hashable) (see also [here](https://docs.python.org/3.9/reference/datamodel.html#object.__hash__)).

By those definitions, our `Edge` objects _aren't_ hashable.

```python
>>> import pydot
>>> e = pydot.Edge("a", "b")
>>> hash(e)
105812016079984353
>>> e.obj_dict["points"] = ("c", "b")
>>> hash(e)
210385293625369463
```

Like the docs say,
> **Most of Python’s immutable built-in objects are hashable; mutable containers (such as lists or dictionaries) are not; immutable containers (such as tuples and frozensets) are only hashable if their elements are hashable.** Objects which are instances of user-defined classes are hashable by default. They all compare unequal (except with themselves), and their hash value is derived from their [id()](https://docs.python.org/3.9/library/functions.html#id).

We get away with putting `Edge` objects in a `set()`, only because it's confined to `to_string` and the Edge's parameters will never change while inside that function. But technically we're violating Python's API by attaching a `__hash__()` to a mutable object, and computing that hash in a way that can be affected by the object's stored values.

PR #263, opened 2021-05-16
Added .exe extension to Dot self.prog when running on standard / pure…

SagaraBattousai:
… windows (i.e. not cygwin).

.exe is appended to self.prog for windows as opposed to assigning by if or in turnary operator to reduce code change required in case self.prog base name changes (i.e. dot to awesome_dot etc)

### DIFF
--- a/pydot.py
+++ b/pydot.py
@@ -1722,6 +1722,11 @@ class Dot(Graph):
 
         self.prog = 'dot'
 
+        # For standard windows (i.e. not cygwin etc) add ".exe"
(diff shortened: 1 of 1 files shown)


ferdnyc:
I think this is handled in the current code, it's changed quite a bit since this PR. From the latest `src/pydot/core.py`:

https://github.com/pydot/pydot/blob/b4500740f8f10b224ac87ce42d972f8480c93d43/src/pydot/core.py#L106-L120

lkk7:
Yes, this is already handled.

PR #267, opened 2021-06-12
Switch to code formatter Black

peternowee:
Proposal for the introduction of the Black code formatter, which promises to save the developers of `pydot` time and mental energy by taking control over code formatting.

See the earlier discussion in pydot/pydot#265 and the individual commit messages in this PR [pydot/pydot#267](https://github.com/pydot/pydot/pull/267/commits).

@ankostis @prmtl @hrishikeshrt Ok, I gave in on the double quotes... but not on the line lengths yet. :smile: Hope you can have a look and let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ start with. Here are 3 common options:
     ```python
     import pydot
 
-    graphs = pydot.graph_from_dot_file('example.dot')
+    graphs = pydot.graph_from_dot_file("example.dot")
     graph = graphs[0]
     ```
 
@@ -81,19 +81,19 @@ start with. Here are 3 common options:
     ```python
     import pydot
 
-    graph = pydot.Dot('my_graph', graph_type='graph', bgcolor='yellow')
+    graph = pydot.Dot("my_graph", graph_type="graph", bgcolor="yellow")
 
     # Add nodes
-    my_node = pydot.Node('a', label='Foo')
+    my_node = pydot.Node("a", label="Foo")
     graph.add_node(my_node)
     # Or, without using an intermediate variable:
-    graph.add_node(pydot.Node('b', shape='circle'))
+    graph.add_node(pydot.Node("b", shape="circle"))
 
     # Add edges
-    my_edge = pydot.Edge('a', 'b', color='blue')
+    my_edge = pydot.Edge("a", "b", color="blue")
     graph.add_edge(my_edge)
     # Or, without using an intermediate variable:
-    graph.add_edge(pydot.Edge('b', 'c', color='blue'))
+    graph.add_edge(pydot.Edge("b", "c", color="blue"))
     ```
 
     Imagine using these basic building blocks from your Python program
@@ -124,14 +124,14 @@ You can now further manipulate your graph using pydot methods:
 - Add further nodes and edges:
 
   ```python
-  graph.add_edge(pydot.Edge('b', 'd', style='dotted'))
+  graph.add_edge(pydot.Edge("b", "d", style="dotted"))
   ```
 
 - Edit attributes of graph, nodes and edges:
 
   ```python
-  graph.set_bgcolor('lightyellow')
-  graph.get_node('b')[0].set_shape('box')
+  graph.set_bgcolor("lightyellow")
+  graph.get_node("b")[0].set_shape("box")
   ```
 
 Output
@@ -155,7 +155,7 @@ Here are 3 different output options:
       the `write_*` methods:
 
       ```python
-      graph.write_png('output.png')
+      graph.write_png("output.png")
       ```
 
 2. Retrieve the DOT string.
@@ -170,7 +170,7 @@ Here are 3 different output options:
       # As a string:
       output_raw_dot = graph.to_string()
       # Or, save it as a DOT-file:
-      graph.write_raw('output_raw.dot')
+      graph.write_raw("output_raw.dot")
       ```
 
     - The Graphviz DOT: You can use it to check how Graphviz lays out
@@ -181,7 +181,7 @@ Here are 3 different output options:
       # As a bytes literal:
       output_graphviz_dot = graph.create_dot()
       # Or, save it as a DOT-file:
-      graph.write_dot('output_graphviz.dot')
+      graph.write_dot("output_graphviz.dot")
       ```
 
 3. Convert to a NetworkX graph.
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 [![Build Status](https://www.travis-ci.com/pydot/pydot.svg?branch=master)](https://www.travis-ci.com/pydot/pydot)
 [![PyPI](https://img.shields.io/pypi/v/pydot.svg)](https://pypi.org/project/pydot/)
+[![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
 
 
 About
--- a/dot_parser.py
+++ b/dot_parser.py
@@ -34,8 +34,8 @@ from pyparsing import (
 
 import pydot
 
-__author__ = ['Michael Krause', 'Ero Carrera']
-__license__ = 'MIT'
+__author__ = ["Michael Krause", "Ero Carrera"]
+__license__ = "MIT"
 
 
 PY3 = sys.version_info >= (3, 0, 0)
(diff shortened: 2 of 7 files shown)


ankostis:
(haven't look at the PR code yet)

As i've written before, double-quotes is like "sealing" the code, guaranteed that no human has touch it since black last run.

I agree the line-length is a choice the developers should be allowed to make, because certain programms may need too frequently longer lines.
I personally like the idea of e -ish flexibilty, though i have not delved into the specifics, when this flexibility kicks-in - not that sometimes when a long line has been split, it will not return back to its short form.

For instance, assuming the next line is too long:

```
foo(a, b, c, d)
```
... then it becomes:
```
foo(
    a,
    b,
    c,
    d,
)
```
... and the last comma(,) prevents it from folding back, even if `b` & `c` are removed:
```
foo(
    a,
    d,
)
```

I guess the same behavior may undo some of the -sh flexibility (if it ever existed in `black`).

ankostis:
- 65988af0.... what an excellent commit, its message and content!  Particularly useful the rejection reasoning for the other tools
- Nice decision to split string-quoting in a separate commit.
- I would run linting early on: it's the most often to fail, and you shouldn't wait for all tests to finish to discover that you forgot it.  But don't know if it should be in its own stage - a failed black should not block further tests.  But that's my own experience.
- Maybe `black --diff` spits out too much output to help the developer locate what has been forgotten?

peternowee:
@ankostis

> ... and the last comma(,) prevents it from folding back, even if `b` & `c` are removed:
> 
> ```
> foo(
>     a,
>     d,
> )
> ```

Yeah, this bothered me as well. There is an option `-C`/`--skip-magic-trailing-comma` now (https://github.com/psf/black/pull/1824) that will make it fold back and remove the comma if it all fits on one line again. But then the surprise is that that also affects manually added magic trailing commas, so there will be no way to manually force explosion to one item per line anymore either. Obviously, `black` cannot tell from a comma if it was placed manually or by `black` itself, so I guess it is one way or the other. Using `-C` actually seems more fitting for Black because it removes the room for discussion about manually adding a magic trailing comma or not. Who knows they might make `-C` the default one day, because things have not been exactly standing still in that area during the last two years already anyway.

> I personally like the idea of e -ish flexibilty, though i have not delved into the specifics, when this flexibility kicks-in - not that sometimes when a long line has been split, it will not return back to its short form.
> [..]
> I guess the same behavior may undo some of the -sh flexibility (if it ever existed in `black`).

If we are talking about the same thing (-ish flexibility == different behavior in the 10% tolerance zone of the maximum line length as discussed in https://github.com/pydot/pydot/pull/265#issuecomment-853290838) then as far as I can tell, that does not exist at all in `black` at the moment. It always uses the configured line length to the max, it is never inclined to break earlier when getting near the maximum and it also never goes over it.

But I think I see what you mean: The way most humans would understand "-ish flexibilty" is that there would be room for the human to decide whether to enter that tolerance zone or not, and then not be corrected by `black`. On the other hand, for `black` "-ish flexibility" would mean a sophisticated ruleset to decide between entering the tolerance zone or breaking the line earlier, but in the end it would still need to be deterministic and disregard what the human decided. Unless there would be a pre-defined way for the human to influence the desired outcome, like these magic trailing commas, but then that would be detrimental to `black`'s goal of not leaving room for discussion again. So maybe indeed the way most humans would interpret "-ish flexibility" is fundamentally incompatible with the goals of `black`.

> * [65988af](https://github.com/pydot/pydot/commit/65988af05062bd21650134bc7087bb63030ab27e).... what an excellent commit, its message and content!  Particularly useful the rejection reasoning for the other tools

Thanks :smile: And thank you for reviewing, btw, I really appreciate it!

> * Nice decision to split string-quoting in a separate commit.

A pleasant side effect of changing my mind on double quotes so late. :sweat_smile:

> * I would run linting early on: it's the most often to fail, and you shouldn't wait for all tests to finish to discover that you forgot it.  But don't know if it should be in its own stage - a failed black should not block further tests, BUT should mark the PR as failing.  But that's my own experience.

Ok, I tried putting the `lint` (`black`) stage first and listing it as `allowed_failure` to prevent it from blocking the `test` stage, but that also meant the entire build would be marked as "Passed" even if the `lint` stage had failed.

So, as you suggest, I moved the job for `black` to the same stage the other jobs use. Now all jobs run almost simultaneously and `black` finishes minutes before the test suite jobs, without blocking them.

Here are two examples using a configuration like that:

- `black` **fail** + test suite **pass**
  https://travis-ci.com/github/peternowee/pydot/builds/228820602
- `black` **pass** + test suite **fail**
  https://travis-ci.com/github/peternowee/pydot/builds/228820429

Both builds are marked as "Failed" by Travis CI.

The change is in the new commit pydot/pydot@9b3d2b7b and I will squash it with the old one pydot/pydot@de074105 at merge time using the commit message of the new one.

> * Maybe `black --diff` spits out too much output to help the developer locate what has been forgotten?

Yeah, I don't know what I can do about that. Dropping the `--diff` would leave them in the dark completely. Where necessary, I suppose we can help them during the discussion of their PR, or if that does not work just make the formatting changes ourselves in a later commit. Maybe just get some experience with this first and then see what we could change or document better.

Also, code changes filed by beginners are usually quite small, right, so then the diff from `black` cannot be very large either. It won't be as bad as the huge diffs we are seeing now for Black's initial runs.

ankostis:
> >  Maybe black --diff spits out too much output to help the developer locate what has been forgotten?
> 
> Also, code changes filed by beginners are usually quite small, right, so then the diff from black cannot be very large either. 

Totally agree.

README.md
```diff
@@ -50,7 +50,7 @@ start with. Here are 3 common options:
     ```python
     import pydot
 
-    graphs = pydot.graph_from_dot_file('example.dot')
+    graphs = pydot.graph_from_dot_file("example.dot")
```
hrishikeshrt, line 53: Double quote changes are okay, I don't have a strong opinion about quotes either way and if anything, uniformity in this regard would be better enforced by `Black`

hrishikeshrt (commented):
Overall, I think this is a very well prepared commit, and will (hopefully) put a rest to further formatting concerns/discussions.

README.md
ankostis: I remind that uniformity in the quotes reduces commit diffs - nowdays there is no code outside of git.

dot_parser.py
```diff
@@ -12,14 +12,25 @@
 import sys
 
 from pyparsing import (
-    nestedExpr, Literal, CaselessLiteral,
-    Word, OneOrMore,
+    nestedExpr,
```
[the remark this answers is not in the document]
peternowee: Agreed. Black does this to minimize diffs, but when it results in a lot of lines it becomes disturbing. The alternative would be to wrap it in `# fmt: off` and `# fmt: on` pragmas, just like I did for the `*_ATTRIBUTES` constants at the top of `pydot.py`. However, I also wanted to keep the number of such exceptions low and thought the problem here (from 8 lines to 19 lines) was not big enough to warrant another exception. Just one of the things we have to accept as the price for switching to `black`, was my conclusion.

dot_parser.py
```diff
@@ -417,84 +443,95 @@ def parse_html(s, loc, toks):
...
         node_stmt = (
-            node_id + Optional(attr_list) +
-            Optional(semi.suppress())).setName("node_stmt")
+            node_id + Optional(attr_list) + Optional(semi.suppress())
+        ).setName("node_stmt")
 
         assignment = (ID + equals + righthand_id).setName("assignment")
-        stmt = (assignment | edge_stmt | attr_stmt |
-                subgraph | graph_stmt | node_stmt).setName("stmt")
+        stmt = (
```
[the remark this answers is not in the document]
peternowee: Yes. Also not big enough to warrant an exception, though.

README.md
peternowee: > uniformity in the quotes reduces commit diffs

As long as people do not touch existing quotes when they make code changes, it should not matter, right? Or you mean when someone makes a code change and at the same time decides to "clean up" the quoting style around that area (as a form of opportunistic refactoring) and then combines it all together in just a single commit? In that case, yes, it would lead to larger diffs and `black` prevents that.

peternowee:
Ok, merged, thanks guys!

PR #274, opened 2021-08-17
add a timeout to avoid hanging forever in some cases

henrywu2019:
In some cases, `dot` will hang and this pydot's `call_graphviz` will never return. To fix it, add a time out to avoid the forever hanging.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -128,7 +128,7 @@ def call_graphviz(program, arguments, working_dir, **kwargs):
         "SYSTEMROOT": os.environ.get("SYSTEMROOT", ""),
     }
 
-    program_with_args = [program] + arguments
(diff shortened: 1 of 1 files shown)


peternowee:
Hi @henrywu2019, thanks for the idea, but a couple of questions that immediately pop up:

- Do you have an example of something that would make `dot` hang? I wonder how much need there is for something like this.
- How do we know what timeout value is long enough? Is 15 seconds going to be enough for everybody?
- At first glance, it looks like your solution depends on a `timeout` command-line program being available. On Debian Linux I hav it, provided by GNU coreutils. But is it also available on Windows? I see Python `subprocess` library has a `timeout` option as well, would that not be better?

Just a few first thoughts. I wonder if we really need this and what complications it might have.

lkk7:
As Peter mentioned, this is probably not portable.

PR #276, opened 2021-09-03
fix output problems when setting non-default progs

kg86:
I found an output problem when setting non-default `prog`s and calling `create_*` or `write_*` functions.
This PR fixes the problem.

The following code reproduces the problem.

Python version: 3.9.6
pydot version: 2.0.0.dev0 (latest)
dot (graphviz) version: 2.43.0

```python
import pydot

dot_string = """graph my_graph {
    bgcolor="yellow";
    a [label="Foo"];
    b [shape=circle];
    a -- b -- c [color=blue];
}"""

graphs = pydot.graph_from_dot_data(dot_string)
graph = graphs[0]
graph.set_prog('neato')
with open('sample1.png', 'wb') as f:
    f.write(graph.create_png())
with open('sample2.png', 'wb') as f:
    f.write(graph.create(format='png'))
graph.write_png('sample3.png')
graph.write('sample4.png', format='png')
```

The outputs are following.

sample1
sample1
sample2
sample2
sample3
sample3
sample4
sample4

I expect that all outputs are the same, but they are not.
`sample1.png` and `sample3.png` (from `create_png` and `write_png`) seem wrong since they do not reflect the attributes `prog='neato'`.

I checked only the single case `prog=neato` and `format=png`, but I guess that the same problems will occur for all other parameters.

If this behavior is intentional, please ignore the PR.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -1591,7 +1591,7 @@ class Dot(Graph):
         # of output in any of the supported formats.
         for frmt in self.formats:
 
-            def new_method(f=frmt, prog=self.prog, encoding=None):
+            def new_method(f=frmt, prog=None, encoding=None):
                 """Refer to docstring of method `create`."""
                 return self.create(format=f, prog=prog, encoding=encoding)
 
@@ -1600,7 +1600,7 @@ class Dot(Graph):
 
         for frmt in self.formats + ["raw"]:
 
-            def new_method(path, f=frmt, prog=self.prog, encoding=None):
+            def new_method(path, f=frmt, prog=None, encoding=None):
                 """Refer to docstring of method `write.`"""
                 self.write(path, format=f, prog=prog, encoding=encoding)
 


ferdnyc (approved):
Hmm. This is interesting. I can reproduce, with current development code. It does seem as though the convenience-method creation routines were accidentally coded to force the initial value of `self.prog` to be their default value, instead of taking the runtime value as they should.

Your change looks like the right approach to fixing that, to me.

I also notice that the docstring for the `create()` method incorrectly claims that its default `prog` is `twopi`, while the `write()` docstring less-incorrectly claims the default is `dot`.

(Really, they should both claim that the default is whatever `self.prog` is currently set to... which initially would be `dot`, unless it's been changed by a call to `set_prog()`.)

cc: @lkk7 — this comes down to a minor mistake in how default args were specified for the `new_method` template functions used to generate all the <code>create_<var>foo</var>()</code> and <code>write_<var>foo</var>()</code> methods. @kg86's fix is simple, effective, and correct.

PR #281, opened 2022-01-04
Update dot_parser.py

blshkv:
https://github.com/pydot/pydot/issues/277


lkk7:
Thanks for the change! It's definitely simpler and I'd prefer this one.
But, unfortunately it didn't solve all the unit tests failing

PR #300, opened 2023-11-14
pydot: support printing nodes without a name

mathstuf:


src/pydot/core.py
```diff
@@ -616,6 +616,8 @@ def to_string(self):
         # RMF: special case defaults for node, edge and graph properties.
         #
         node = quote_if_necessary(self.obj_dict["name"])
+        if node is None:
+            node = "<unnamed>"
```
ferdnyc, line 620: Hmmm. I'm not sure this is a good idea, only because multiple unnamed nodes being given the _same_ name will cause problems.

Consider the difference between this:
```dot
graph G {
  A [color=red]
  B [shape=box]
}
```
and this:
```dot
graph G {
  <unnamed> [color=red]
  <unnamed> [shape=box]
}
```

The first renders as:
image

the second renders as:

image

Really, it's an error to have an unnamed node, and we should probably complain about that sooner. (Raise an exception when `pydot.Node(None, ...)` is called, for instance. There's an argument to be made that the user could call `set_name()` on it _later_, granted. But, really I think the code is already too permissive, and should be less shy about treating errors as errors instead of doing its best to save the user from themselves.)

But if we really _do_ want to protect users from themselves in this situation, I think we'd have to generate a unique name each time a nameless node is encountered.

ferdnyc: One thing we _could_ do, I just realized — because we have the `sequence_number` which is unique for each entity in the graph — is generate a unique name from _that_ for unnamed nodes.

There's still the issue that any generated string might conflict with the user's own names, but I think if we used something sufficiently wacky we'd be pretty safe from collisions.

Perhaps:
```suggestion
        if node is None:
            node = f"__unnamed__node__{self.get_sequence()}__"
```

TBH I'm sort of on the fence about this. I'm still kind of partial to actually just _letting_ a nameless node be an error condition, or even yelling **harder** at the user, if they try to generate source for a graph with an unnamed node.

Maybe we could raise a more _explicit_ error directly, but having a node with no name is a nonsensical condition that _**should**_ cause an error. I'm worried that generating a name might be doing more harm than good, by hiding the user's own bugs from them.

cc: @lkk7, WDYT?

ferdnyc: Bleah! Except, `get_sequence()` could _also_ return `None`, if the Node hasn't been added to a Graph yet.

The more I think about it, the less I like the idea of creating node names ourselves when the user didn't bother.

mathstuf: I'm trying to remember what I was doing here…I feel I was analyzing the output of `ninja -t graph` via `python-graph-diff`. These graphs apparently just have unnamed nodes (not sure why else I would have found this). The uniqueness of the name didn't matter much in my case (apparently). I'm having trouble finding the actual code I was using :/ .

lkk7: Example graph from ninja: https://graphviz.org/Gallery/directed/ninja.html

lkk7: Wait, we can't get have unnamed nodes in graphviz. We can only create them in pydot. So that graph is irrelevant.

I think we should either:
- Error out
- Raise a warning, and proceed with a default unique name

But the first option is simpler.

ferdnyc: And arguably more correct, because it's not like the user can create any edges or anything, to a node with an autogenerated ID. I almost feel like we'd be doing the user a disservice by creating them.

ferdnyc: > it's not like the user can create any edges or anything, to a node with an autogenerated ID

Well... okay, _technically_ they COULD, with pydot:

```python
n1 = pydot.Node(None, color="red" label="West")
n2 = pydot.Node(None, color="blue" label="East")
e = pydot.Edge(n1, n2)
```

...But that implies we'd have to generate the node names _immediately_ after they're created, since the `Edge` constructor is going to expect to get a result from `n1.get_name()` and `n2.get_name()`. And again, because all of that can happen (like above) **before** anything is added to a graph and assigned a sequence number, we can't even rely on those to generate the names. I suppose we could just assign them UUIDs, or something... :face_with_diagonal_mouth:

ferdnyc: @mathstuf 

> I'm trying to remember what I was doing here…I feel I was analyzing the output of `ninja -t graph` via `python-graph-diff`. These graphs apparently just have unnamed nodes (not sure why else I would have found this). The uniqueness of the name didn't matter much in my case (apparently). I'm having trouble finding the actual code I was using :/ .

Your fork of python-graph-diff has an [`unlabeled-nodes` branch](https://github.com/mathstuf/python-graph-diff/tree/unlabeled-nodes), perhaps that holds some clues?

As @lkk7 pointed out, Graphviz itself **can't have** unlabeled nodes, so if this need arose from some situation where (presumably valid) graphs were being used as _input_ to pydot, then I'm inclined to say the existence of unnamed nodes is likely a bug in either graph-diff or some other code. The original Ninja graph (like any Graphviz graph) can't _possibly_ have had unnamed nodes, since in reality there's just no such thing. A node is **entirely** defined by its ID.

ferdnyc: (That being said, I'd still support raising an exception if `None` is passed to either the first argument of `pydot.Node()` or either of the first two arguments to `pydot.Edge()` (unless the first argument is a list/tuple<sup>1</sup>). But not _creating_ names, since I think we've fully covered all of the unavoidable ways that will certainly go horribly wrong at least some of the times.)

### Notes
1. Though I _really_ hate that feature and would love to make it go away, TBH. Accepting _both_ of the endpoints in the first argument to `pydot.Edge()` as a list needlessly complicates the code (and complicates the _typing_ of that code, even harder), and I can't think of any valid reason for it. The only _invalid_ reason I can think of is that maybe some of the `dot_parser` code was written too lazily and works that way, but if that's the case I'd rather just fix THAT code to call the constructor correctly.

mathstuf: I forget if that `get_label()` was returning `None` or not…probably was as I don't think it'd have raised an issue before. However, the diff is wrong as `get_labe` isn't a thing, so how much testing did I really do of this "fix"…?

lkk7:
I closed this because I think we know enough now not to support unnamed nodes.
Instead, https://github.com/pydot/pydot/issues/400 has been created.

PR #296, opened 2023-03-31
Fix multiple breaking issues from new `pyparsing` versions

lkk7:
This PR:
- Fixes broken whitespace behavior which caused almost every dot file to be parsed incorrectly (https://github.com/pydot/pydot/issues/291, https://github.com/pydot/pydot/issues/289, https://github.com/pydot/pydot/issues/280, https://github.com/pydot/pydot/issues/277)
- Fixes broken HTML parsing (found through tests `html_labels.dot`, `html2.dot`)
- Fixes broken parsing for identifiers with non-standard characters (found through test `russian.dot`)
- Removes a few unused imports/variables.
- Bumps the version of `pyparsing` to `>=3.0.9`

Tradeoff: from my tests, there is a small performance degradation due to introduction of `pyparsing_unicode.BasicMultilingualPlane.alphanums` which impacts `pyparsing`'s speed. This can be investigated later. For now, the most important thing is that it works.

PR #302, opened 2023-12-16
Add GitHub Actions CI, modernize packaging

ferdnyc:
This PR makes several changes to the repo, and might be more easily reviewed commit-by-commit, though there are a lot of them so perhaps it's best to just go file-by-file.

* Packaging is modernized
  * Python 3.7 as the minimum version
  * The bulk of metadata, and all dependency configs, moved from `setup.py` to `pyproject.toml`
  * Unsupported `setup.py` feature `tests_require` migrated to a new optional dependency set defined by the `tests` extra (`pydot[tests]`)
  * Other extras defined in `pyproject.toml`:
    * The old `dev` extra, containing `chardet` and `black`
    * A `release` extra, which installs the `zest.releaser` tool along with its `[recommended]` addons

* The `pyparsing` dependency now _requires_ a version `<3`, due to the issues with pyparsing 3.0+. This allows pydot to install and run on all platforms. @lkk7's PR #296 should be updated, if this is merged first, to then also remove that restriction.

* Testing is scaffolded
  * `tox` is now the testing framework, with its configuration defined in `setup.cfg`. (there's no `tox.ini`, another config file seemed unnecessary)
  * `tox` is configured with test environments for every Python version 3.7–3.12 inclusive (`py37` to `py312`)
  * Black linting is also made part of the `tox` configuration, with an additional environment `black` defined to run the tool under Python 3.12
  * The means by which `test/pydot_unittest.py` discovers the path to its own directory is updated (see diff).

* GitHub Actions CI is enabled (building on the work started in #299)
  * A separate test instance will be run for each Python 3.7 – 3.12
  * Currently tests are only run on `ubuntu-latest` platform, but MacOS and/or Windows can be added if we wish
  * Black linting is performed as part of, and only under, the Python 3.12 testing, in the same runner.
  * [toxdev/tox-gh](https://github.com/toxdev/tox-gh) is used to integrate with the GitHub Actions matrix, and on each runner instance will run only the set of environments defined for the installed Python.

* All of Black's changes to the source formatting are applied and checked in, so that the existing code passes its linting.

* The Travis and AppVeyor configs are deleted. (They're in the gi t history if anyone ever needs them.)

...I think that's about it. Have at it! Any reviews/feedback welcome.


ferdnyc:
A Successful run of the GH Actions CI config can be seen here:

https://github.com/ferdnyc/pydot/actions/runs/7231448851

ferdnyc:
RE: The GH Actions permissions, when running workflows in my fork I noticed that the `GITHUB_TOKEN` had full read/write permissions to the repo. I'm not sure if that config was defaulted generally, or copied from the upstream repo (here), but for security reasons it's advisable to lock it down.

It'd be helpful if someone with repo admin rights could go into **Settings > Actions > General** and change the "Workflow permissions" configuration (bottom of the page) from "Read and write permissions" to "Read repository contents and packages permissions", if it's not already selected.

Nothing the CI is doing requires repo-write permissions. The only reason that would need to change is if we decided to have Black automatically apply its own corrections.

ferdnyc:
> * Unsupported `setup.py` feature `tests_require` migrated to a new optional dependency set defined by the `tests` extra (`pydot[tests]`)

I'm not sure it's actually necessary to have `chardet` and `black` in there, since they need to be installed to the _inner_ environments `tox` creates — and they're redundantly defined as dependencies in the `setup.cfg` toxenv sections for that reason. Installing them in the _outer_ Python is actually unnecessary/pointless.

But, it makes `pydot[tests]` a superset of `pydot[dev]`, and means users won't have to install _both_ extras separately if they want the full testing/dev environment.

lkk7:
@ferdnyc Thanks for this great work, I see that tests pass and there will probably be no reviewers as of now.
I'm ready to merge this if you don't plan on adding anything else

ferdnyc:
@lkk7 Nope, good to go from my perspective! Only outstanding item is whether we want macOS/Windows CI as well.

ferdnyc:
(Even if we do, that can be a followup PR.)

lkk7:
The original CI had some Windows workflow, and it was something conda-related. I don't know conda that much to know if it's still relevant. That can be looked at in the future changes.

ferdnyc:
It's a party in pydot.

image

lkk7:
Great, I'll look at my change and make sure it's OK. And update the pyparsing version

pyproject.toml
```diff
@@ -1,3 +1,60 @@
...
+requires = ["setuptools"]
+
+[project]
+name = 'pydot'
+description = "Python interface to Graphviz's Dot"
+readme = "README.md"
+license = {file = "LICENSE"}
+requires-python = ">= 3.7"
+dependencies = [
+  'pyparsing>=2.1.4,<3'
```
ferdnyc, line 11: @lkk Yeah, you just need to drop the `,<3` here, in #296.

PR #307, opened 2023-12-18
Update MANIFEST.in for better sdist contents

ferdnyc:
A quick pre-2.0.0 tweak (if I made it in time).

When running `prerelease` from `zest.releasing`, its check-manifest scan registers some mismatches between the repo and the sdist. These changes to the `MANIFEST.in` ensure that the check passes, and the sdist contents are exactly what we/zest expect.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,10 @@
 include README.md
 include LICENSE
 include ChangeLog
-include requirements.txt
-include test/*_unittest.py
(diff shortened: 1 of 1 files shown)


ferdnyc:
> (if I made it in time)

...I did not.

lkk7:
I think you did, there's no release yet

lkk7:
Thank you!

PR #308, opened 2023-12-18
Massive unit testing overhaul: Parameterization, parallelization, and objects, objects, objects!

ferdnyc:
This PR makes, incrementally over a series of commits, sweeping changes to the way the project unit tests are structured, executed, and analyzed. 

**TL;DR**: Unit tests are more granular, and execute in parallel for faster testing.

For the way-too-long detailed version, read on.

### Parameterization

Instead of the output-comparison tests in `test/my_tests` and `test/charts` being executed as one long test case that chugs through all of the files in  a given directory, the `parameterize` PyPi module is now used to construct a _set_ of test cases, one per dotfile, and run each one individually.

The overall effect is, the test case count increases from 33 to 233, with no change in runtime, but much more granular output from the unittest run. Instead of the entire set of regression tests looking like this:
```text
test_graphviz_regression_tests (__main__.TestGraphAPI.test_graphviz_regression_tests)
 ...#######################################################3
```
...as it grinds through 197 files in that directory, now those files' tests will be output individually, like so:
```text
  test_regression_001_rootlabel (test.test_pydot.TestGraphvizRegressions.test_regression_001_rootlabel)
 ... ok
  test_regression_002_labelroot_fbc (test.test_pydot.TestGraphvizRegressions.test_regression_002_labelroot_fbc)
 ... ok
  test_regression_003_ldbxtried (test.test_pydot.TestGraphvizRegressions.test_regression_003_ldbxtried)
 ... ok
  test_regression_004_labelclust_ntc (test.test_pydot.TestGraphvizRegressions.test_regression_004_labelclust_ntc)
 ... ok
  test_regression_005_pm2way (test.test_pydot.TestGraphvizRegressions.test_regression_005_pm2way)
 ... ok
  test_regression_006_b80a (test.test_pydot.TestGraphvizRegressions.test_regression_006_b80a)
 ... ok
  test_regression_007_colors (test.test_pydot.TestGraphvizRegressions.test_regression_007_colors)
 ... ok
  test_regression_008_labelclust_fbl (test.test_pydot.TestGraphvizRegressions.test_regression_008_labelclust_fbl)
 ... ok
  test_regression_009_labelroot_ndc (test.test_pydot.TestGraphvizRegressions.test_regression_009_labelroot_ndc)
 ... ok
```
(The tests are auto-numbered, the rest of the identifier string is based on the input dotfile name.)

### Parallelization

The previous change was about making the tests more granular without impacting performance. Now we get to take advantage of that increased granularity for some massive performance gains.

The `unittest-parallel` tool is a test runner compatible with the standard-library `unittest` module, but as its name suggests it can execute tests in parallel. Now that we have 233 quick tests to break up, instead of just 33 (one of which accounted for over 50% of the total runtime all by itself), we can divide those 233 cases up and execute them in parallel, making use of all cores on the test system.

In my testing on a local four-core Intel Core i5 (that's over 10 years old), total test execution runtime dropped by over 2/3, from `~95s` to just over `30s` total.

On the GitHub runners the effects of this are negligible; with a hot cache it might shave 10 seconds off runtime for each runner, and those times were already hovering around the 60-second mark. But if more, or more complex, tests are added, the impact could be greater.

### Class hierarchy restructuring

As part of making these changes, I reworked how the test cases are structured, breaking up the single monolithic `unittest.TestCase` subclass into several.

* The parameterized tests are each their own class, one per directory.
* The tests that run (or don't) on the files in `test/from-past-to-future` are also their own class, and the logic to execute or skip that test based on the presence of the directory now uses the `@unittest.skipUnless()` decorator to evaluate that precondition.
* All of the tests that compare rendered results from `pydot` to the output of the `dot` command are now subclasses of `RenderedTestCase` (which defines the `_render_and_compare_dot_file()` method they use)
* And _everything_ is a subclass of `PydotTestCase`, which defines the common `_reset_graphs()` method all classes need.

Separately, there are now two non-testcase classes, `Renderer` and `RenderResult`, related to those `RenderedTestCase` test. 

`Renderer` is a non-instantiated class that holds the former `_render_with_pydot()` and `_render_with_graphviz()` methods (now called `Renderer.graphviz()` and `Renderer.pydot()`), and those both return their output in an instance of `RenderResult`.

`RenderResult` is instantiated with the result data (the image bytes) from the render method, and it handles the creation of the `sha256().hexdigest()` internally.

### Backwards-compatibility kludges

After I put all this together, I discovered that Python 3.8 and lower lack `str.removesuffix()`, and Python 3.7 is also missing `functools.cached_property()`.

I use `str.removesuffix()` to strip the `".dot"` extension off the end of the filenames for the parameterized tests, so if that's not available I just leave the suffix on.

But `cached_property` is what makes the `RenderResult` class efficient, since it only has to compute its `sha256.hexdigest` once for the stored data, and even then only on-demand. I didn't want to give that up, so I ended up vendoring just the `cached_property` class from the Python 3.12 `functools.py` into the `test/` directory as `_functools.py`, that way if it's missing from the standard-library `functools` module I can just use `from ._functools import cached_property` instead.

One unfortunate side effect of that is that the test module now _has_ to be run as a package module, not as a Python script. That only applies to Python 3.7, fortunately, but in that version the relative import of `._functools` will fail when trying to execute the tests with `python test/test_pydot.py`. (That's the script's new name, more on that momentarily.) Running `unittest` or `python -m unittest` with _no_ arguments, however, will successfully run the tests, so it's easier to just do that. (Or even better, use `tox`, which knows how to run the tests correctly and in parallel.)

### Discoverable tests

The reason for the rename from `test/pydot_unittest.py` to `test/test_pydot.py` is that the latter name is compatible with the test discovery built in to both unittest and unittest-parallel. When run with no arguments, they recursively scan python packages in the local directory looking for test modules, and one of the default discovery patterns is `test_*.py`. There's also a (completely empty) `test/__init__.py` file now, for the same reason — discovery won't look inside a directory that's not a python package.

Since unittest-parallel doesn't support being handed a script name as an argument, and since the tests can't be run in parallel from the script's `if __name__ == '__main__'` block, it just works out better all around if the unit tests are discovered automatically instead of executed manually.


ferdnyc:
Oh, yes, and the `black` scan performed by `tox` now covers the entire repo, including `test/`, instead of only looking in `src/`.

lkk7 (approved):
That's a potential whole blog post with a lot of useful information 😄 I've planned to refactor the tests later but you've done it in a much better way.
Are you able to check it on Windows?

ferdnyc:
@lkk7 What's the supported build environment there? (Either way the answer is yes, I have a local Win10 system with Visual Studio 2019 and MSYS2 both installed...)

lkk7:
To be honest, I'm not sure. I've never used it on Windows

ferdnyc:
Oh, right. It's a pure Python project and that question makes exactly ZERO sense. Neeeevermind. (It's really late.) :wink:

lkk7:
@ferdnyc Thanks for the refactor. Looks like things are working as expected:
- Failures on Windows
- Failing a single unit test doesn't stop the whole run.
From my perspective, it's ready to merge.

ferdnyc:
@lkk7 I would concur.

The more granular tests are also informative and interesting. Windows runners show anywhere from 2-5 falling tests, and it's not at all clear whether those counts are related to Python version, are random/racy, or what. Getting more runs of these should be helpful.

PR #309, opened 2023-12-18
Reenable two disabled tests

ferdnyc:
Two of the tests in `test/graphs/` had names not ending in `.dot`, indicating they were disabled because they caused crashes in the `dot` program. That was over 10 years ago, and testing today my local `dot` has no problems with them at all. Let's reenable them and see if anything breaks.


ferdnyc:
Huh. I guess they really do crash `dot` on Ubuntu. (Where, in 22.04, the `dot` is from graphviz 2.43.0.)

On Fedora with dot from graphviz 8.1.0, `b15.dot` outputs this message, but doesn't crash:
```console
$ dot -Tjpg < test/graphs/b15.dot > /tmp/b15.jpg
Warning: flat edge between adjacent nodes one of which has a record shape - replace records with HTML-like labels
  Edge HoverForwardToStop -> HoverToFlight
```
...And `b58.dot` has no problems at all, not even a warning.

...Still, we're testing on Ubuntu, not Fedora, so I guess it's best to leave these disabled for now.

PR #310, opened 2023-12-19
Add Windows and macOS CI

ferdnyc:
The Windows runs currently have some failing tests, so they're configured to fail without aborting the workflow.

(I'll have to check whether the Windows failures interfere with PR mergeability. They may need to be adjusted to actually fake-pass, instead of simply ignoring the failures.)

#### Windows failing tests
There are a few files in `test/charts/` that Windows `dot.exe` (from graphviz 9.0.0, installed via Chocolatey) fails on. 

1. `b51.dot` is the one that shows up in the current CI. (It's the first failure in that directory, and the big monolithic test case stops there.) It fails with a hexdigest mismatch — `dot.exe` just seems to render it a bit differently.
1. `b53.dot`, same as `b51` just renders differently and triggers a digest mismatch.
1. `proc3D.dot` is a third digest mismatch.
1. `b993.dot` is different, it never actually renders at all. The graph data starts with the line `diGraph G{`, and that triggers a syntax error. It seems that Windows `dot.exe` is case-sensitive, ironically, so it doesn't recognize `diGraph` as a synonym for `digraph`.

(Note: One effect of merging #308 would be that all tests are run and every failure reported, rather than skipping all the rest of the tests in the directory at the first sign of failure.)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,11 +21,16 @@ concurrency:
 
 jobs:
   tests:
-    name: "Test with Python ${{ matrix.python-version }}"
-    runs-on: "ubuntu-latest"
+    name: "${{ matrix.system.os }}, Py${{ matrix.python-version }}"
+    runs-on: "${{ matrix.system.os }}-latest"
+    continue-on-error: ${{ matrix.system.can-fail }}
 
     strategy:
       matrix:
+        system:
+          - {os: ubuntu, can-fail: false}
+          - {os: windows, can-fail: true}
+          - {os: macos, can-fail: false}
         python-version:
           - "3.12"
           - "3.11"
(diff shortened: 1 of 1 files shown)


ferdnyc:
I've been thinking it might be a good idea to dump the failing-test comparison images into their own directory somewhere, instead of into the current dir. (Even if we ran the tests from an empty dir, tox pollutes its own working directory pretty badly so it's not the best place to store output.)

My basic thought was to have the CI workflow export whatever's in the failure-output dir as build artifacts, IFF it finds something in there.

lkk7 (approved):
Thanks! Yeah, I remember some things failing on Windows a few months ago when I've done my change.

The case sensitivity is interesting.
> The keywords node, edge, graph, digraph, subgraph, and strict are case-independent.
(https://graphviz.org/doc/info/lang.html)

Is it a bug to submit?

Ironically, I fell ill just as I gained access to the repository, so I haven’t been able to review much. But I do trust you if you think either of the two PRs is ready to merge. This doesn't doesn’t impact the actual PyDot code anyway.

ferdnyc:
@lkk7 

> The case sensitivity is interesting.
> 
> > The keywords node, edge, graph, digraph, subgraph, and strict are case-independent.
> > (https://graphviz.org/doc/info/lang.html)
> 
> Is it a bug to submit?

I don't see any other way to interpret it, so I've opened https://gitlab.com/graphviz/graphviz/-/issues/2481 with a minimum reproducible example.

ferdnyc:
@lkk7 

So, IMHO it makes sense to merge this, with the Windows tests failing (but configured as allowed-to-fail). Once it's merged, we can tackle trying to fix the broken Windows tests. (Even if that means just configuring those individual tests to be skipped on Windows.)

Merging this would also provide cross-platform CI testing for #308, which would hopefully provide a path towards confidently merging that PR.

Issue #306, opened 2023-12-16
Rare inconsistencies between Graphviz and Pydot


lkk7:
In unit tests which compare hashes between pydot-generated and graphviz-generated images, there may (rarely) occur small inconsistencies. 

Example:
![err_clust2 dot_graphviz_18_06_03](https://github.com/pydot/pydot/assets/14022447/41279a09-5f98-4491-851a-9ba15648fd51)
instead of:
![err_clust2 dot_pydot_18_06_03](https://github.com/pydot/pydot/assets/14022447/8948e2e0-2b4a-452d-8d22-791085195d53)

They might randomly break unit tests and they need to be investigated.

More details needed. Latest PR makes failing tests like these get saved, so I'll be able to look at them when they occur.

ferdnyc:
The outcome of the upstream investigation (see [graphviz issue 2510](https://gitlab.com/graphviz/graphviz/-/issues/2510) was that this is due to a bug in graphviz, recently fixed in [graphviz MR 3487](https://gitlab.com/graphviz/graphviz/-/merge_requests/3487).

The fixes are included in the recently-released graphviz 10.0.1, so anyone using the latest version shouldn't see these problems anymore.

PR #316, opened 2024-02-12
CI: Export failed images as build artifact

ferdnyc:
- Make saving of images from failed tests conditional on setting an envvar `TEST_ERROR_DIR` to a path in which to store them.
- Modify image storage to use predictable names inside uniquely named subdirs.
- Provide a script to create image diffs with ImageMagick.
- Export failed-test dirs and comparison script as an artifact from failed CI runs.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -30,7 +30,7 @@ jobs:
         system:
           - {os: ubuntu, can-fail: false}
           - {os: windows, can-fail: true}
-          - {os: macos, can-fail: false}
+          - {os: macos, can-fail: true}
         python-version:
           - "3.12"
           - "3.11"
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -47,6 +47,15 @@ jobs:
         with:
           python-version: "${{ matrix.python-version }}"
           cache: "pip"
+      - name: "Set up TEST_ERROR_DIR"
+        shell: bash
+        run: |
+          mkdir test_errors
+          if [ "${{ runner.os }}" = "Windows" ]; then
+            echo "TEST_ERROR_DIR=$(cygpath -w $(realpath ./test_errors))" >> "$GITHUB_ENV"
+          else
+            echo "TEST_ERROR_DIR=$(pwd -P)/test_errors" >> "$GITHUB_ENV"
+          fi
       - name: "Install graphviz (Linux)"
         if: runner.os == 'linux'
         run: sudo apt install graphviz
@@ -64,4 +73,13 @@ jobs:
         run: tox -vv --notest
       - name: "Run tests"
         run: tox --skip-pkg-install
-
+      - name: "Prepare error artifacts"
+        if: ${{ failure() }}
+        run: cp test/compare_images.sh "${{ env.TEST_ERROR_DIR }}"
+      - name: "Publish error artifacts"
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: test-errors-${{ runner.os }}-${{ matrix.python-version }}
+          path: ${{ env.TEST_ERROR_DIR }}
+          if-no-files-found: ignore
--- a/test/compare_images.sh
+++ b/test/compare_images.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+for dir in */; do
+    magick composite $dir/err_pydot.jpeg \
+        -compose difference $dir/err_graphviz.jpeg \
+        $dir/err_difference.png
+done
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,6 +20,8 @@ env_list =
 extras = tests
 package = wheel
 wheel_build_env = .pkg
+pass_env =
+    TEST_ERROR_DIR
 commands = unittest-parallel --level test -vv
 
 [testenv:black]
--- a/test/test_pydot.py
+++ b/test/test_pydot.py
@@ -24,6 +24,7 @@ if sys.version_info >= (3, 8):
 else:
     from ._functools import cached_property
 
+TEST_ERROR_DIR = os.getenv("TEST_ERROR_DIR", None)
 
 TEST_PROGRAM = "dot"
 TESTS_DIR_1 = "my_tests"
@@ -94,6 +95,25 @@ def _load_test_cases(casedir):
         return fname.removesuffix(".dot")
 
     return [(_case_name(dot_file), dot_file, path) for dot_file in dot_files]
+
+
+def _compare_images(fname: str, pydot: RenderResult, gv: RenderResult) -> bool:
+    """Compare two RenderResult objects for the named test.
+
+    If the images differ and a ``TEST_ERROR_DIR`` has been provided, create
+    a subdir for the test and dump both images there for examination."""
+    if pydot.checksum == gv.checksum:
+        return True
+    if TEST_ERROR_DIR is not None:
+        dirname = fname.replace(".", "_")
+        out_dir = os.path.join(os.path.normpath(TEST_ERROR_DIR), dirname)
+        os.makedirs(out_dir)
+        pydot_path = os.path.join(out_dir, "err_pydot.jpeg")
+        gv_path = os.path.join(out_dir, "err_graphviz.jpeg")
+        with open(pydot_path, "wb") as p, open(gv_path, "wb") as g:
+            p.write(pydot.data)
+            g.write(gv.data)
+    return False
 
 
 class PydotTestCase(unittest.TestCase):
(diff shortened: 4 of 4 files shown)


.github/workflows/CI.yml
```diff
@@ -39,6 +39,9 @@ jobs:
           - "3.8"
           - "3.7"
 
+    env:
+      TEST_ERROR_DIR: "$RUNNER_TEMP/test_errors"
```
lkk7, line 43:
```suggestion
      TEST_ERROR_DIR: "${{ runner.temp }}/test_errors"
```
Looks like something's wrong this variable.
https://github.com/pydot/pydot/actions/runs/7869461894/job/21468622150#step:9:508
Will this fix the problem?

ferdnyc: Yeah, sorry, I'd meant to mark this as draft as it's not there yet. I can fix _that_ problem with `${{ env.RUNNER_TEMP }}` (or your suggestion, I'm sure), but the mismatched directory slashes on Windows still foul things.

test/compare_images.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+for dir in */; do
+    magick composite $dir/err_pydot.jpeg \
```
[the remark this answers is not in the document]
ferdnyc: Nope, this is never run by the CI system. It's just a helper script that'll be included in the artifacts zip, to be run by the downloader if they're so inclined. (I didn't want to waste the CI time both installing and running ImageMagick, as well as having to deal with the whole Windows/non-Windows scripting "thing" of actually doing the `magick` calls.)

.github/workflows/CI.yml
ferdnyc: (I added an `os.normpath()` to the `test_python.py` script, which fixes the directory slashes for the test run... but then they're still wrong when setting the `paths` for `actions/upload-artifact`. Frickin' Windows.)

ferdnyc:
This might do it.

ferdnyc:
Wait, what!? Do the CI jobs not run for draft PRs??? UGH.

ferdnyc:
Oh, nope, it's that I broke the workflow syntax.

.github/workflows/CI.yml
ferdnyc: Grrr. I'm beginning to suspect that `runner.temp` isn't defined yet, when setting envvars. I reduced it to just this, and still no go:

```yaml
env:
  TEST_ERROR_DIR: ${{ runner.temp }}
```

ferdnyc:
Grrr. Mostly got it, but macOS has no `realpath` (unless we go through the hassle of doing a `brew install coreutils`), so I had to work around that.

ferdnyc:
So, here's an animated GIF of the differences in `b51.dot`, in the latest Python 3.12 @ Windows run. That turned out to be a better comparison method than the ImageMagick-composited difference blend, for this particular output.

(It's a big GIF, so this may look like crap.)
err_b51

ferdnyc:
I feel like I should've attached an epilepsy warning to that. :roll_eyes:

lkk7:
Very interesting differences.
This one fails always, not randomly, right?

Then it's probably not connected in any way to #306 
(which already has an issue reported https://gitlab.com/graphviz/graphviz/-/issues/2510)

ferdnyc:
> Very interesting differences. This one fails always, not randomly, right?

On Windows, AFAICT yes. It's the most consistent of the failures. Whatever else is listed among the failing tests, `b51.dot` will be there.

It also _sometimes_ fails on macOS, as well. (macOS with Python 3.7, at least. I think I need to set `can-fail: true` for the Mac jobs also, so we can get a better picture of which versions fail, and when.

The macOS and Windows `dot`-generated versions seem to have identical layout, modulo the fact that the images are different sizes, different aspect ratio, and use different fonts.

The other weird thing is, the hashes — either from pydot or graphviz — are almost never the same, from different runs. Which strikes me as odd — _at least_ the `err_graphviz.jpeg` files should be identical every time, I'd think! It's the same `dot.exe` binary producing all of them, after all.

I suppose that could just be lossy JPEG artifacting at play. I wonder if rendering to PNG might produce more consistent results?

(In the most recent set of runs, the Windows Python 3.11 and Python 3.12 `err_graphviz.jpeg` files have the same hash, but all the others are different.):

```console
$ sha256sum Win*/b51_dot/err_graphviz.jpeg |cut -d/ -f1
b20e36b55408a8ace0d0a59e75b3c1b5b563cb8b361c983affb4cec21e32bc1a  Win310
8fda742fc7c68565bb23b299936c017226c912876f828ca4d35216db27cd0dad  Win311
8fda742fc7c68565bb23b299936c017226c912876f828ca4d35216db27cd0dad  Win312
f92aec990ca922355a92557e21959276d574bd56e4aa78a09eecef1289b8079b  Win38
e5153a47a2c34d80b19e909cda14f760f0d447f2ff8c99d8968b43e78a0d69d7  Win39

$ sha256sum Win*/b51_dot/err_pydot.jpeg |cut -d/ -f1
3dceb97cf6450e3046f1977e5c8fd89c3a4c261115464ddb011ed6ad94be9398  Win310
4dffa41cd22b2a90074a3993ac305c93c261c6a7c17f985051fb0a51d7990c13  Win311
dbb9450ea2bbc7ae22f81cf7ecde9dd4ae8807cd02c887d52b2ab3e5d11bed67  Win312
1b5e03299d5bec8bf725b173a1fc8d8fdb3c9c65307df19d9c8ce1bad92cda93  Win38
9a00a9d472ec105ee6a7c7550d45e2248e6dbd58940a72fe486a532ef9b93eed  Win39
```

ferdnyc:
Here's the animated GIF of the Win-py310 vs. Win-py312 `b51_dot/err_graphviz.jpeg` differences. (With a lower framerate, this time.) 

b51_win310_win312


Very similar to the previous pydot-vs-graphviz animation for Win-py312. So, clearly it's not the JPEG compression, it's that on Windows this graph seems to flip-flop between at least two different layouts seemingly at random. But... consistently-random(??), somehow? (Enough that the layout in pydot always seems to be the opposite of what `dot.exe` produced.)

ferdnyc:
I guess it's not that surprising that the JPEGs differ, since `dot -Tdot b51.dot` produces a differently-laid-out graph definition in each of: 

* Fedora's graphviz-8.1.0 `dot`
* my locally-built graphviz-10.0.0~dev.20231228.2101 `dot`
* Windows `dot.exe` from chocolatey-installed graphviz-9.0.0.

lkk7:
> I guess it's not that surprising that the JPEGs differ, since `dot -Tdot b51.dot` produces a differently-laid-out graph definition in each of:
> 
>     * Fedora's graphviz-8.1.0 `dot`
> 
>     * my locally-built graphviz-10.0.0~dev.20231228.2101 `dot`
> 
>     * Windows `dot.exe` from chocolatey-installed graphviz-9.0.0.

It may be best to disable this test, as well as any others that fail due to graphviz differences or random fluctuations in it. Testing graphviz is pointless, right? I know the `can_fail: true` is there for Windows but I've noticed it still "makes the CI results red".

Also, side note: https://gitlab.com/graphviz/graphviz/-/issues/2510 says that some random problems should have been fixed in the latest versions (which I have trouble installing on Mac)

ferdnyc:
@lkk7 Ah, I hadn't seen that, and in fact just opened https://gitlab.com/graphviz/graphviz/-/issues/2514 after discovering that four runs of `dot.exe -Tjpeg b51.dot -o b51.jpg` on Windows produced four different files.

But I agree, and have just added a note to this effect, that this may be a consequence of the same (now fixed) bug.

I believe I can test the latest versions on both Windows and macOS, so I'll try running a version of `dot` built with graphviz PR !3487 included, and see if it now produces consistent results.

ferdnyc:
I've now confirmed that `dot.exe` from graphviz-10.0.1 on Windows renders `b51.dot` consistently the same with every run, so maybe we disable these tests for now (I'm _guessing_ the other failures are the same bug), and reenable them when Chocolatey's graphviz gets updated to 10.0.1.

ferdnyc:
Anyway, as far as _this_ PR, it's now working as expected, so taking it back out of Draft status. From my POV it's ready to merge. Disabling `b51.dot` and friends can be a separate PR.

.github/workflows/CI.yml
ferdnyc: ^ They weren't, so I switched to adding a step early in the job run that sets `TEST_ERROR_DIR` instead.

test/compare_images.sh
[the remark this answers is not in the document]
ferdnyc: (It's somewhat annoying that GitHub strips execute permissions from anything included in the artifact zip —  though I understand their motivations, from a security perspective — so the script has to be run with `sh ./compare-images.sh` or after doing a `chmod +x`.)

ferdnyc:
> Disabling `b51.dot` and friends can be a separate PR.

In fact, it's #317.

PR #317, opened 2024-02-14
Disable certain tests temporarily on Windows

ferdnyc:
These tests render unpredictably in graphviz `dot.exe` until version 10.0.1. See https://gitlab.com/graphviz/graphviz/-/issues/2510

(Note: I've made this Windows-only for now, as that's the platform where tests fail _reliably_. macOS is exhibiting **occasional** failures as well, but we don't have enough CI runs yet to be sure which tests fail there.

(Also, I expect Homebrew will be updated to graphviz 10.0.1 long before Chocolatey will, so the macOS problem should solve itself soon enough.)

### DIFF
--- a/test/test_pydot.py
+++ b/test/test_pydot.py
@@ -392,6 +392,8 @@ class TestGraphAPI(PydotTestCase):
         self.assertIsInstance(pydot.__version__, str)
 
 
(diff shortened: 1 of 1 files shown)


ferdnyc:
\o/ Green across the board, that's what we like to see.

PR #318, opened 2024-02-14
Move method generation to class level

ferdnyc:
Instead of generating methods in the `__init__()` of the class instances, generate them as class modifications immediately after the class definition. This allows better introspection and type checking.

* In the interactive REPL, typing `pydot.Node.<tab>` will show all of the generated attribute setters and getters.
* Code editors will similarly display valid completions on strings such as `pydot.Node.set_` or `pydot.Dot.write_`.
* Mypy's `stubgen` or `pyright --createstub` can generate type stubs that include the generated methods, with the correct commandline flags applied.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -81,6 +81,46 @@ CLUSTER_ATTRIBUTES = {
 # fmt: on
 
 
+OUTPUT_FORMATS = {
+    "canon",
+    "cmap",
+    "cmapx",
+    "cmapx_np",
+    "dia",
+    "dot",
+    "fig",
+    "gd",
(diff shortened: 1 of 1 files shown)


ferdnyc:
CI failure is an apparently-spurious layout issue in one of the macOS runners, #319 disables that test on macOS in addition to Windows.

ferdnyc:
Huh. And I just discovered that eerily-similar changes are also part of @peternowee's PR #242, tho that's unmerged after more than three years so I think calling it "stalled" would be an understatement. Perhaps making the changes in smaller increments, like with this PR, will help shepherd them through the review process.

PR #320, opened 2024-02-16
Fix quoting of attribute values

### LINKED ISSUE #301: Style is wrongly represented in the dot file for  style="rounded,filled" elements
pydot has problem with rendering dot file that would contain "rounded,filled" elements (which is supported by graphviz).
In pydot it's not encapsulated in "" sign causing rendering error.

Sample python file to reproduce it.

`import pydot

# Create the graph
graph = pydot.Dot(graph_type='digraph', fontname="Verdana", compound=True)

# Set the graph attributes
graph.set_node_defaults(shape="box", style="rounded", fontname="Verdana", margin="0.3,0.2")
graph.set_edge_defaults(fontname="Verdana")
graph.set_rankdir('LR')
graph.set_remincross(True)
graph.set_pack(True)

# Add the nodes
login = pydot.Node("Login", label="Login \n 100%", shape="doublecircle", fillcolor="cornflowerblue", color="black",
                   style="rounded,filled", fontcolor="black")
nodes = ["platformA", "platformB"]
node_attrs = [
    {"label": "platformA", "fillcolor": "gray", "fontcolor": "black", "style": "rounded,filled", "color": "black"},
    {"label": "platformB", "color": "black", "fillcolor": "seagreen", "style": "rounded,filled", "color": "black"},
]

for node, attr in zip(nodes, node_attrs):
    graph.add_node(pydot.Node(node, **attr))

# Add the subgraphs
cluster_logins = pydot.Subgraph(graph_name="cluster_logins", graph_type='digraph', color='black',
                                fillcolor='cornflowerblue', label='Logins', rank='same', style='rounded,filled')

logout = pydot.Node("logout", label="logout", fillcolor="cornflowerblue", color="black", style='rounded,filled',
                    fontcolor="black")

cluster_logout = pydot.Subgraph(graph_name="cluster_logout", graph_type='digraph', style='rounded,filled',
                                fillcolor='tan2', label='logout')

# Add the nodes to the subgraphs
cluster_logins.add_node(pydot.Node("platformA"))
cluster_logins.add_node(pydot.Node("platformB"))

cluster_logout.add_node(logout)

# Add the edges
graph.add_edge(pydot.Edge(login, "platformA", color="gray"))
graph.add_edge(pydot.Edge(login, "platformB", color="seagreen"))

# Add the subgraphs to the main graph
graph.add_subgraph(cluster_logins)
graph.add_subgraph(cluster_logout)

graph.add_edge(pydot.Edge("platformA", logout, color="seagreen"))
graph.add_edge(pydot.Edge("platformB", logout, color="seagreen"))

# Write the graph to a DOT file
graph.write_raw("reproduce.dot")
graph.add_node(login)
# Add the other nodes
`

ferdnyc:
The `needs_quotes()` and `quote_if_necessary()` logic for handling numeric values and/or values containing commas seems to have been written based on a comma-decimal numeric format, leaving strings like `"1,234"` unquoted while forcing quotes around strings like `"1.234"`.

From a pydot perspective, that's exactly backwards, since attributes are comma-separated so all values containing commas should be quoted. This also bled into things like `style` values containing multiple classes (e.g. `"rounded,filled"`), which would be left unquoted.

This PR replaces the `'[0-9,]*'` numeric regular expressions with `r'[0-9\.]*'`, and equivalent for the alphanumeric regexps, so that values like `"1.234"` are left unquoted, but `"1,234"` and `"rounded,filled"` will always be quoted.

A second commit adds a couple of additional unit tests to verify the new behavior.

Fixes #301

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -285,11 +285,11 @@ class frozendict(dict):
 
 dot_keywords = ["graph", "subgraph", "digraph", "node", "edge", "strict"]
 
-id_re_alpha_nums = re.compile("^[_a-zA-Z][a-zA-Z0-9_,]*$", re.UNICODE)
+id_re_alpha_nums = re.compile(r"^[_a-zA-Z][a-zA-Z0-9_\.]*$", re.UNICODE)
 id_re_alpha_nums_with_ports = re.compile(
-    '^[_a-zA-Z][a-zA-Z0-9_,:"]*[a-zA-Z0-9_,"]+$', re.UNICODE
+    r'^[_a-zA-Z][a-zA-Z0-9_\.:"]*[a-zA-Z0-9_\."]+$', re.UNICODE
 )
-id_re_num = re.compile("^[0-9,]+$", re.UNICODE)
+id_re_num = re.compile(r"^[0-9\.]+$", re.UNICODE)
 id_re_with_port = re.compile("^([^:]*):([^:]*)$", re.UNICODE)
 id_re_dbl_quoted = re.compile('^".*"$', re.S | re.UNICODE)
 id_re_html = re.compile("^<.*>$", re.S | re.UNICODE)
@@ -314,8 +314,12 @@ def needs_quotes(s):
     if s in dot_keywords:
         return False
 
-    chars = [ord(c) for c in s if ord(c) > 0x7F or ord(c) == 0]
-    if chars and not id_re_dbl_quoted.match(s) and not id_re_html.match(s):
+    has_high_chars = any([ord(c) > 0x7F or ord(c) == 0 for c in s])
+    if (
+        has_high_chars
+        and not id_re_dbl_quoted.match(s)
+        and not id_re_html.match(s)
+    ):
         return True
 
     for test_re in [
--- a/test/test_pydot.py
+++ b/test/test_pydot.py
@@ -299,6 +299,26 @@ class TestGraphAPI(PydotTestCase):
             self.graph_directed.get_nodes()[0].to_string(), "node [shape=box];"
         )
 
+    def test_comma_separated_attribute_values_to_string(self):
+        self._reset_graphs()
+        self.graph_directed.add_node(
+            pydot.Node("node", color="green", style="rounded,filled")
+        )
+        self.assertEqual(
+            self.graph_directed.get_nodes()[0].to_string(),
+            'node [color=green, style="rounded,filled"];',
+        )
+
+    def test_attribute_string_values_quoting(self):
+        self._reset_graphs()
+        self.graph_directed.add_node(
+            pydot.Node("node", length=1.234, size="2.345", radius="9,876")
+        )
+        self.assertEqual(
+            self.graph_directed.get_nodes()[0].to_string(),
+            'node [length=1.234, radius="9,876", size=2.345];',
+        )
+
     def test_names_of_a_thousand_nodes(self):
         self._reset_graphs()
         names = {"node_%05d" % i for i in range(10**3)}


src/pydot/core.py
```diff
@@ -314,8 +314,12 @@ def needs_quotes(s):
...
         return False
 
-    chars = [ord(c) for c in s if ord(c) > 0x7F or ord(c) == 0]
-    if chars and not id_re_dbl_quoted.match(s) and not id_re_html.match(s):
+    has_high_chars = any([ord(c) > 0x7F or ord(c) == 0 for c in s])
+    if (
+        has_high_chars
+        and not id_re_dbl_quoted.match(s)
+        and not id_re_html.match(s)
+    ):
```
ferdnyc, line 322: Not strictly necessary as part of this change, but IMHO more readable, and definitely more efficient. Instead of storing a list of the ordinal value for every non-ASCII character in the string, just store a single boolean indicating whether it contains any non-ASCII characters (since that's all we care about).

ferdnyc:
CI failure is a flaky test failing on macOS, #321 changes the configs to disable all of them on both non-Linux platforms.

lkk7:
Hi,
I haven't had much time to look at this one, but is it possible that this breaks something for European locales?
I mean, it's good that it doesn't quote 1.234, and it's good that it quotes "rounded,filled", but why should it quote "1,234" if it's just a number? We'll, it's ambiguous, because it could also mean 1 followed by 234, which are two attributes that should be quoted.
So your change is for the better and I'm ready to merge it, just curious.

ferdnyc:
@lkk7 
> I mean, it's good that it doesn't quote 1.234, and it's good that it quotes "rounded,filled", but why should it quote "1,234" if it's just a number? We'll, it's ambiguous, because it could also mean 1 followed by 234, which are two attributes that should be quoted.

That's exactly it — even if `1,234` is a floating-point value, because attributes are always comma-separated that number would need to be quoted (seems to me) to be distinct from an attribute with the value `1`, followed by an attribute named `234`. Same way `style="filled,rounded"` needs to be quoted, to not be parsed as `style=filled`, `rounded`.

IMHO it would be fine for it to continue quoting `1.234` as well (like it's been doing until now) when it's a string, but since it _won't_ quote it when it's an actual `float` value, it makes sense not to quote it when it's a string, either.

As far as Euro locales, I admit I don't know enough about how tat works (a) in Python, and (b) in graphviz, to answer definitively.

If I start up Python with `LC_ALL=de_DE` in the environment, it still uses `.` as the decimal separator, and `,` as the list item separator. It's only when I import `locale` and explicitly set the numeric locale to e.g. `de_DE`, that it starts using comma decimals in `locale.format_string()` — but the REPL still uses `.`.

```python3
 $ LC_all=de_DE LC_NUMERIC=de_DE python3
>>> x=1,234
>>> x
(1, 234)
>>> x=1.234
>>> x
1.234
>>> import locale
>>> locale.format_string("%f", x)
'1.234000'
>>> locale.getlocale(locale.LC_NUMERIC)
(None, None)
>>> locale.setlocale(locale.LC_NUMERIC, "de_DE")
'de_DE'
>>> locale.format_string("%f", x)
'1,234000'
>>> x=1,234
>>> x
(1, 234)
>>> x=1.234
>>> x
1.234
```

So, long and short of it:

1. It's not clear to me whether comma decimals are ever recognized in Python code, outside of the processing done by the `locale` module.
1. Even if they are, given the `node_id [attr_id=value, attr_id=value]` format used by graphviz, it seems like numerical values containing commas would always need to be quoted, in any locale, to — as you say — avoid parsing ambiguity.

Though I certainly wouldn't turn down input from any users in comma-decimal locales, regarding how this needs to be handled.

PR #324, opened 2024-02-20
Tests: Reenable flaky tests on Windows & macOS

ferdnyc:
graphviz-10.0.1 is now available through Chocolatey & Homebrew.

(I know [I predicted](https://github.com/pydot/pydot/pull/317#issue-2134839664) Chocolatey would lag behind Homebrew and be the last ones to update, when it turned out the other way around. That's because I put my finger on the scale: [I reported](https://github.com/chocolatey-community/chocolatey-packages/issues/2432) the package as out-of-date to Chocolatey, _and_ [I fixed](https://gitlab.com/graphviz/graphviz.gitlab.io/-/merge_requests/653) the sorting issues on the download page that were hiding the new release way at the bottom of the list, so both systems' auto-updaters were able to find the new release.)

Regardless, they've both now gotten there, and 10.0.1 should contain the fixes that make our flaky tests no longer flaky, so reenabling them on both platforms.

### DIFF
--- a/test/test_pydot.py
+++ b/test/test_pydot.py
@@ -432,11 +432,6 @@ class TestGraphAPI(PydotTestCase):
         self.assertIsInstance(pydot.__version__, str)
 
 
-# TODO: Remove skipIf when graphviz 10.0.1 is available
-@unittest.skipIf(
-    sys.platform.startswith(("win", "darwin")),
-    "Unreliable on Windows and macOS",
-)
 class TestShapeFiles(PydotTestCase):
     shapefile_dir = os.path.join(_test_root, "from-past-to-future")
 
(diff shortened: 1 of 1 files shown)


ferdnyc:
Hm. I need to `black` my code, but regardless the Windows installers started downloading 9.0.0 still.

I've [triple-checked](https://community.chocolatey.org/packages/Graphviz) that 10.0.1 is really, truly available in Chocolatey, so the issue must be caching.

It just showed up _today_, so maybe it's a matter of just waiting a few hours, or maybe we have to explicitly `choco install graphviz --version 10.0.1` at least once to kick the cache.

ferdnyc:
OK, explicitly specifying the version number did work to force an install of 10.0.1.

ferdnyc:
And using 10.0.1 does fix the flaky tests on Windows!

Now let's see if we can drop the explicit `--version 10.0.1`; I don't want to leave it there and miss _newer_ releases down the line.

ferdnyc:
Grrr. Py3.7, Py3.8, and Py3.9 still installed 10.0.1.
Py3.10, Py3.11, and Py3.12 went back to 9.0.0.

ferdnyc:
Eh, I'll just leave it alone for now and try again tomorrow, there's no rush.

lkk7:
I re-ran the CI and the result is the same. Should the explicit version be set? Or maybe there's an option to specify "at least this version" `>=`?

ferdnyc:
> I re-ran the CI and the result is the same. Maybe there's an option to specify "at least this version" `>=`?

Not that I can find, nor is there any flag to force a refresh AFAICT. Which seems nuts, but 🤷.

If it's still happening (I'll try another run momentarily), I suppose we can leave the explicit version in there for a week or two, then try pulling it out again. Hopefully by that time everything will have caught up.

ferdnyc:
All Windows **_and_ all macOS versions** are now pulling 10.0.1, looks like I can remove the disable checks completely.

PR #327, opened 2024-03-13
Modernize from-past-to-future.dot (so that it renders properly)

ferdnyc:
For some time now, the images in the graph haven't been rendering. That's because:

- Using `style=invis` today prevents the node from rendering at all (I've opened an issue about the fact that this is apparently a change, and undocumented: https://gitlab.com/graphviz/graphviz/-/issues/2522)
- `shapefile` is deprecated in favor of `image` https://www.graphviz.org/faq/#FaqCustShape
- `shape=box` is redundant with `shapefile`, as it's the only option
- With `image` nodes (like any other node), `shape=plaintext` specifies no outline

The changes here create a graph with embedded images that have no outlines, seemingly as the original authors intended.

### DIFF
--- a/test/from-past-to-future/from-past-to-future.dot
+++ b/test/from-past-to-future/from-past-to-future.dot
@@ -10,55 +10,55 @@ digraph G {
 
 	/* Nodes */
 	subgraph cluster_Computers {label="Computers"; labelloc="b"; Computers_icon};
-	Computers_icon [label="", shape=box, style=invis, shapefile="Computers.png"];
+	Computers_icon [label="", shape=plaintext, image="Computers.png"];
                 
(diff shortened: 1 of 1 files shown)


ferdnyc:
#### Before

from-past-to-future

#### After

new-fptf

ferdnyc:
I discovered that `shape=none` (a) exists, and (b) seems like it communicates intent better than `shape=plaintext`. They appear to be completely equivalent.

lkk7:
Some detailed research on GitLab is going on, I see! 
Is this ready to merge?

ferdnyc:
@lkk7 Yep. Whatever the vagaries of `style="invisible"` vs. `style="invis"` turn out to be, from-past-to-future needn't use either of them. `shape=none` is the more sensible way to not draw nodes, which is what this code now does.

PR #328, opened 2024-03-14
Replace all remaining %-substitution and .format() calls with f-strings

ferdnyc:
With Python 3.6 and higher, there are few reasons to ever use `%`-substitution or explicit `.format()` calls, and lots of reasons to prefer `f`-strings instead.

(When Python 3.7 support is dropped, we could even take advantage of the final improvement older Python versions lack: [the `=` specifier](https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging), which allows for self-documenting string generation/output:

```python
>>> my_variable = 27.4
>>> print(f"{my_variable=}")
my_variable=27.4
```
...Which would likely be useful in `test_pydot.py`, if not the main package code. But 3.7 doesn't support that syntax.)

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -18,7 +18,7 @@ except Exception as e:
     warnings.warn(
         "`pydot` could not import `dot_parser`, "
         "so `pydot` will be unable to parse DOT files. "
-        "The error was:  {e}".format(e=e)
+        f"The error was:  {e}"
     )
 
 
@@ -280,7 +280,8 @@ class frozendict(dict):
             return h
 
     def __repr__(self):
-        return "frozendict(%s)" % dict.__repr__(self)
+        dict_repr = dict.__repr__(self)
+        return f"frozendict({dict_repr})"
 
 
 dot_keywords = ["graph", "subgraph", "digraph", "node", "edge", "strict"]
@@ -457,8 +458,8 @@ def graph_from_adjacency_matrix(matrix, node_prefix="", directed=False):
             if e:
                 graph.add_edge(
                     Edge(
-                        "%s%s" % (node_prefix, node_orig),
-                        "%s%s" % (node_prefix, node_dest),
+                        f"{node_prefix}{node_orig}",
+                        f"{node_prefix}{node_dest}",
                     )
                 )
             node_dest += 1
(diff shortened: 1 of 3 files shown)


src/pydot/core.py
```diff
@@ -1784,25 +1781,15 @@ def create(self, prog=None, format="ps", encoding=None):
...
-                arguments=arguments,
-                code=process.returncode,
-                out=stdout_data,
-                err=stderr_data,
+            code = process.returncode
+            print(
+                f'"{prog}" with args {arguments} returned code: {code}\n\n'
+                f"stdout, stderr:\n {stdout_data}\n{stderr_data}\n"
             )
-            print(message)
```
ferdnyc, line 1797: Sections like this really display the benefits `f`-strings have over previous methods. 10 lines of code become 4.

ferdnyc:
> the final improvement older Python versions lack: [the `=` specifier](https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging)

Whoops! Guess not, I'm behind the times. [Python 3.12 added a bunch of `f`-string improvements](https://docs.python.org/3/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings), including backslash escapes and comments _inside_ replacement fields:

```python
>>> line1 = "This is a message."
>>> line2 = "It should be newline-separated."
>>>
>>> # Legal since Python 3.6
>>> print(f"{line1}\n{line2}")
This is a message.
It should be newline-separated.
>>>
>>> # Only legal since Python 3.12
>>> print(f'{"\n".join([line1, line2])}')
This is a message.
It should be newline-separated.
>>>
>>> # Also only legal since Python 3.12
>>> print(f'{line1}\n{line2 # Second line}')
This is a message.
It should be newline-separated.
```

ferdnyc:
Oh, and one more: The parsing context inside replacement fields is no longer shared with the enclosing `f`-string itself, which has implications for reuse of quote characters.

Basically, _this_ is now legal (if confusing) in Python 3.12:

```python
>>> print(f"{"\n".join([line1, line2])}")
```

There isn't much to recommend _that_ particular construct, vs. the outer-single-quoted, Python 3.6+-compatible version. But for more complex `f`-strings, it could help avoid some ugliness or forced escaping.

lkk7:
Yeah, the code needs some solid linting and typing

ferdnyc:
@lkk7 I've slowly been making inroads on typing. #318 was a big first step — since the introspection system couldn't even _see_ the generated methods before, it certainly couldn't type-check them.

The next thing I need to do is get `frozendict` specified correctly, since this is completely un-typeable:

```python
class frozendict(dict): 
    def _blocked_attribute(obj):
        raise AttributeError("A frozendict cannot be modified.")
                          
    _blocked_attribute = property(_blocked_attribute)
                         
    __delitem__ = __setitem__ = clear = _blocked_attribute
    pop = popitem = setdefault = update = _blocked_attribute
```

Once frozendict has properly-typed interface methods that override the standard `dict` methods, then its uses can be type-checked accordingly.

Issue #118, opened 2016-04-01
Automatially quote node names containing a colon


josch:
Hi,

if a node name contains a colon, then an interesting thing happens:

```python
>>> g = pydot.Dot()
>>> g.add_node(pydot.Node('foo:bar'))
>>> print(g.to_string())
digraph G {
foo;
}
```

As you can see, the node is now named `foo` and not `foo:bar`. I can see elsewhere that this might be called a "feature" of pydot. Unfortunately, one of the more popular users of pydot, networkx, sees this as a bug instead:

https://github.com/networkx/networkx/issues/2050

I want to clarify with you what exactly is the case from your point of view? Is this a pydot bug or a pydot feature? If the former, can it get fixed?

johnyf:
Thanks for noting this. The relevant place where the port name after a colon is removed from a node's name is [`pydot` L760](https://github.com/erocarrera/pydot/blob/ebfd155d60a6000fedb2169c0d963ed5acfcf0e1/pydot.py#L760).

From reading the code, my understanding is that this behavior duplicates GraphViz's behavior. The rationale appears to be to detect and collect port names automatically from node names, and prepare node names by dropping the port name.

A small experiment shows that not changing the node's name (perhaps still detecting port names automatically) produces the same result after layout with GraphViz. Therefore, the end result (image) will be the same.

Unlike the image, `Dot.to_string` won't be the same as before. It will be the string that the user gave as node name, instead of a string modified by `pydot`. Based on:

- the principle of least surprise
- PEP20: flat better than nested (modifications outside `pydot` easier than fixed ones inside)
- PEP20: explicit better than implicit (silent string modifications are implicit, and become visible only after the first failure, or after reading the entire docs of GraphViz)
- my previous experience that a `pydot`-`dot2tex` toolchain with each tool making small modifications to strings results in a combination that is very difficult to use

I prefer to not modify the strings that a user defines as node names. Ports could be detected automatically (as currently done), but that is implicit. Explicit is better, and since this is Python and not text input to GraphViz, it is reasonable to have the user define ports as node attributes, separately (as much work, or less, than appending a `:port_name` to a string).

The above is under the assumption that I am not missing interference with anything else (though I would still favor this change). Moreover, it does not change the API. If ports are defined separately, then `pydot` should append itself the `:port_name` automatically, which is probably what a user would expect more (also along the lines of how the `networkx` API works with attributes).

peternowee:
Still reproducible with pydot 1.4.1.

[...]

PR #339, opened 2024-04-30
Fix quoting logic for attribute values

### LINKED ISSUE #258: String attributes containing colon are not quoted
```
import pydot
n = pydot.Node('x', fillcolor='red:yellow')
print(n)
```

Seen:

```
x [fillcolor=red:yellow];
```

Expected:

```
x [fillcolor="red:yellow"];
```

(`dot` won't compile without the value being quoted, complaining of `Error: file.dot: syntax error in line x near ':'`)

This behavior has been confirmed on pydot 2.0.0.dev0.

ferdnyc:
This started off as an effort to simply fix quoting for attribute values containing colons (#258), which was being hosed by the logic that recognizes ID values in the form <code><var>foo</var>:<var>n</var></code> as IDs with an optional port value.

That's what made me realize that the rules for quoting IDs and the rules for quoting attributes are totally different, and we shouldn't be trying to make them the same.

So, this PR breaks out `quote_if_necessary()` into two functions, `quote_id_if_necessary()` (which keeps the same logic as before), and `quote_attr_if_necessary()`, which is _heavily_ biased towards quoting. It'll basically quote anything that isn't numeric, HTML, or already double-quoted. Simple string values like `shape=box` that used to be left unquoted will now be output as `shape="box"`, because it's never wrong to do that. And it avoids a lot of problems.

I also removed the `re.UNICODE` flag from all the regexps, as in Python3 that's a tautology, and replaced a `'"' + s + '"'` with `fr'"{s}"'`. (Though that needn't have been a raw string, come to think of it.)

Three tests (only 3!!) had to be adjusted, so their expected values matched the new quoting rules.

Fixes #258

(Doesn't fix the other issues involving IDs, like #118)


ferdnyc:
Hmm. It seems GitHub has started running some `macos:latest` jobs on Apple Silicon machines, and Python for darwin-arm64 didn't show up until 3.8.10.

Python 3.7 is actually [EOL as of 2023-06-27](https://devguide.python.org/versions/#unsupported-versions), so we _could_ just drop it entirely.

ferdnyc:
Set to Draft state because I just noticed there's [a full grammar for ID strings](https://graphviz.org/doc/info/lang.html) in the Graphviz docs, including regular expressions, which don't match ours. Seems like a good idea to make them match.

ferdnyc:
So, turns out I was _mostly_ right. The distinction isn't between attribute values and IDs, but between IDs, and `node_ID`s when they're used to define nodes or edges.

All attribute values _are_ IDs, in their grammar, but only the `node_id` in a `node_stmt`<sup>1</sup> or on either side of an `edge_stmt` can have ports and compass directions. So, the magic logic we've been using to detect `:`-separated ID values should _only_ apply to node IDs. For any other type of ID, colons are just another reason to double-quote.

Technically we could even double-quote node IDs, since all IDs are just strings and it makes no difference to Graphviz whether we write...

```dot
# This
strict digraph {
    a [shape=ellipse style=filled fillcolor="#1f77b4"]
    b [shape=polygon style=filled fillcolor="#ff7f0e"]
    a -> b [fillcolor="#a6cee3" color="#1f78b4"]
}

# Or this
strict digraph {
    "a" [shape="ellipse" style="filled" fillcolor="#1f77b4"]
    "b" [shape="polygon" style="filled" fillcolor="#ff7f0e"]
    "a" -> "b" [fillcolor="#a6cee3" color="#1f78b4"]
}
```

We just have to be careful that...

```dot
# This... (OK)
digraph {
  node1:port1 -> node2:port5:nw;
}

# Doesn't get turned into this... (WRONG)
digraph {
  "node1:port1" -> "node2:port5:nw";
}

# But it can be turned into this... (OK)
digraph {
  "node1":"port1" -> "node2":"port5":"nw";
}
```

Part of me almost thinks we should give the user the ability to set the `port` and `compass_pt` _separately_ from the node name, to avoid any ambiguity about whether their colons are separators, or just colons in the name.

IOW, if a user had to define the graph above not with this:

```python
G = pydot.Dot()
G.add_edge(pydot.Edge("node1:port1", "node2:port5:nw"))
```

But with something like _this_:

```python
G = pydot.Dot()
node1 = pydot.Node("node1")
node2 = pydot.Node("node2")
ep1 = pydot.Endpoint(node1, port="port1")
ep1 = pydot.Endpoint(node2, port="port5", compass="nw")
G.add_edge(pydot.Edge(ep1, ep2))
```
Then constructing the Dot syntax would be a whole hell of a lot easier. And if we provided some method of configuring auto-port/compass parsing _off_ for users who don't want it, they could name their nodes things like `"01: Math"` and `"02: Science"` without having to worry about them possibly being misinterpreted.

Right now, we attempt to _parse out_ the `port` from a `Node()` name, but we don't allow those values to be set unambiguously. We even have a `get_port()` for `Node` objects, but no `set_port()`!

Even worse, if you set your graph up like this:
```python
G = pydot.Dot()
node1 = pydot.Node("node1:port1")
node2 = pydot.Node("node2:port5:nw")
G.add_edge(pydot.Edge(node1, node2))
```
You'll get this output:
```dot
digraph G {
node1 -> node2;
}
```
We lose everything except the name, when a `Node` object is used as an `Edge` endpoint.

And that `node2`? Here's how we parse it:
```python
>>> node2.get_port()
':port5:nw'
```
Whoops.

Maybe I'll work on adding `set_port()`, `get_compass()`, and `set_compass()` to `Node`s (with fixed parsing), and an `Endpoint` class as an optional no-ambiguity method of specifying complex edges.

Heck, we could even provide an enum for the valid compass points, to be used like:

```python3
ep1 = pydot.Endpoint(
    node2, port="port5", compass=pydot.Compass.nw)
```

### Notes

1. The graphviz grammar _says_ that node statements start with `node_id` values, so I'm taking it on faith that it's useful to create and output node statements like this:
    ```dot
    digraph {
      node2:port5:nw [label="Node2!", color="red"];
      node1:port1 -> node2:port5:nw;
    }
    ```
    But I'm not really clear on why that would ever be useful, and I don't see any actual examples of it in their docs so far. Afaict it's completely identical to this:
    ```dot
    digraph {
      node2 [label="Node2!", color="red"];
      node1:port1 -> node2:port5:nw;
    }
    ```
    I've [posted a question to the Graphviz forum](https://forum.graphviz.org/t/ports-compass-points-in-node-statements/2163) about that.

ferdnyc:
> 1. I've [posted a question to the Graphviz forum](https://forum.graphviz.org/t/ports-compass-points-in-node-statements/2163) about that.

So, no less than Stephen North hisself [has confirmed](https://forum.graphviz.org/t/ports-compass-points-in-node-statements/2163/2?u=ferdnyc):


```dot
#Complex Node statements like:
nodeName:portname:nw [attributes=values];

# Are meaningless, and always equivalent to...
nodeName [attributes-values];
```

He says it was a "small mistake", that the grammar was specified to make the former legal syntax. The `node:port:compass` form was only _intended_ to be used for edge endpoints.

Obviously we can't just disregard the grammar entirely, but IMHO (and as I argued in that discussion), I feel this _does_ give us the leeway to parse `Node` calls like:

```python3
pydot.Node("01:Math", color="red")
```

As if the user actually wrote:
```python3
pydot.Node("\"01:Math\"", color="red")
```

And we can therefore assume they expect _this_ as the resulting `.dot` code:

```dot
"01:Math" [color=red];
```

As opposed to the original, un-double-doublequoted `pydot.Node()` call defining a node named `01` with a port named `Math`, which is how we currently interpret it.

lkk7:
That's a long journey.

BTW the port definition is weird, there's another port in form of "<...>", but in a different context (https://graphviz.org/doc/info/shapes.html – section "Record-based nodes").

ferdnyc:
> That's a long journey.
> 
> BTW the port definition is weird, there's another port in form of "<...>", but in a different context (https://graphviz.org/doc/info/shapes.html – section "Record-based nodes").

Yeah, that goes inside the label attribute for the destination node, it's how you provide targets for the edge-endpoint ports. Technically without those labels, any ports attached to endpoints do nothing; _with_ labels defined, the `node:port` syntax can target individual cells of the record-based nodes.

Apparently that whole "thing" is now discouraged in favor of [HTML-like labels](https://graphviz.org/doc/info/shapes.html#html-like-label-examples) anyway. (Which can also define ports, by adding a <code>port="<var>name</var>"</code> attribute to target HTML entities.)

lkk7:
(Closing and reopening to re-trigger CI)

lkk7:
Hi @ferdnyc,

I was wondering if you'll be continuing the work on this one, or should I try to take it over.

ferdnyc:
@lkk7 Ooh, thanks for the reminder! I'd totally forgotten about this. Let me resolve the conflicts and see what state it's currently in, then I'll figure out where to go from there.

lkk7:
Thanks, that's great because you've already been deep into this issue and I'm sure you'll do it better.

ferdnyc:
So, turns out there's basically **nothing** here that hasn't already been handled in subsequent, more-focused PRs. Closing this as no longer relevant. (Yay!)

PR #340, opened 2024-04-30
Drop Python 3.7 from CI matrix

ferdnyc:
It's EOL as of 2023-06-27, and causes problems for MacOS jobs running on Apple Silicon, for which there is no version available.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,7 +40,6 @@ jobs:
           - "3.10"
           - "3.9"
           - "3.8"
-          - "3.7"
 
     steps:
       - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # v4.1.2


lkk7:
Hi @ferdnyc, thanks for another batch of good ideas, I'll look at all those PRs soon.
You sure you don't want to join the maintainers? No strings attached (I did send you an email about this once, but it might have been considered spam).

ferdnyc:
@lkk7 
> You sure you don't want to join the maintainers? No strings attached (I did send you an email about this once, but it might have been considered spam).

Never saw that, sorry! Spamcanned is one possibility. Another, equally-likely possibility is that it's one of my... \*checks\*... 6,169 unread messages. :roll_eyes: 

...I usually just tell people that email is Not A Good Way To Reach Me™. :grin: 

Sure, I'll pitch in. I pinky-swear promise not to [install any backdoors into pydot](https://www.wired.com/story/xz-backdoor-everything-you-need-to-know/).

PR #344, opened 2024-05-04
Fix my embarrassing typo

ferdnyc:
Apparently I can't spell "MIT". But hey, I got 2 of the 3 letters right, so that's something?


ferdnyc:
(And I can't spell "embarrassing", either!)

lkk7 (approved):

Issue #352, opened 2024-06-02
Create hard parameters for graphviz attributes?


2br-2b:
When creating a Graphviz graph, there isn't any documentation for the parameters built into the system, which means there's no intellisense and no way to see what parameters exist or what their valid values are.

The currently best way for a beginner is to go to the official documentation for graphviz (for example, https://graphviz.org/docs/edges/) to see the possible values and set them there, but it would be easier to have built-in enums. For example, here's how I am creating my edges:

```python
class ArrowStyle(Enum):
    FORWARD = "forward"
    BACKWARD = "back"
    BOTH = "both"
    NONE = "none"

class JwEdge(pydot.Edge):
    def __init__(self, source_id: str, destination_id: str, arrow_direction: ArrowStyle = ArrowStyle.NONE):
        if label == None:
            label = ""
        
        super().__init__(
            src=source_id,
            dst=destination_id,
            dir=arrow_direction.value,
            )
```

On one hand, this makes it easier for me to create edges; on the other hand, I don't know how static Graphiz's api is, so having an attribute there that might not be supported for a specific version could lead to other issues.

Related to #130

lkk7:
Yeah, that's a difficult thing to solve.
I also like Intellisense and typing wherever possible, but keeping up with all the attributes, possible values, and types is an unrealistic effort for me now.
And as you said, there's the problem of having different versions of Graphviz (even though it's probably very stable now, you never know what's going to happen or what weird version someone will use).

ferdnyc:
> And as you said, there's the problem of having different versions of Graphviz (even though it's probably very stable now, you never know what's going to happen or what weird version someone will use).

This isn't just a theoretical concern, either. There are extant cases where [graphviz takes undocumented or conflictingly-documented values for certain attributes](https://gitlab.com/graphviz/graphviz/-/issues/2525), which create a real difficulty for PyDot to handle sensibly.

ferdnyc:
The other tricky thing here is, if this were to be implemented, the corresponding getters and setters would really need to be defined using the enums as their output/input types (respectively). Whereas, currently, all of those methods are generated from lists of names (and, of course, therefore generated _untyped_) at class definition time.

PR #353, opened 2024-06-02
Make the formats into an enum rather than a list of strings

2br-2b:
What would be your thoughts on some sort of change like this? This would need some work before being merged, but this way, intellisense can detect supported formats?

A change like this specific one would be a breaking change, so this is probably not the best way to implement this (maybe make the enum a separate class and generate OUTPUT_FORMATS based on this), but I just want to get thoughts before I spend too much work on something like this

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -10,6 +10,7 @@ import subprocess
 import sys
 import tempfile
 import warnings
+from enum import Enum
 
 import pydot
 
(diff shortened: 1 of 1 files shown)


2br-2b:
Related to #352

lkk7 (commented):
Thanks for the PR!

As you said, it can be a breaking change. For anyone using the raw `write` or `create` method. I don't think many people do that, but they might.

Maybe the solution could be: `create` and `write` accept either `str` or enum so that their API is not broken. 
Enum would be encouraged, and used internally, but `str` would also work and be converted to enum instantly.

(Another thing: right now the tests fail because in some tests **strings** are passed to `create`, instead of the now expected **enum**).

src/pydot/core.py
```diff
@@ -1681,7 +1684,7 @@ def write(self, path, prog=None, format="raw", encoding=None):
                 f.write(s)
         return True
 
-    def create(self, prog=None, format="ps", encoding=None):
+    def create(self, prog=None, format=OUTPUT_FORMATS.PS, encoding=None):
```
ferdnyc, line 1687: Wow... unrelated, but we should change that default. Who wants PostScript!?

ferdnyc:
@2br-2b 
> but this way, intellisense can detect supported formats?

I guess that answers my "why?", but I'm not sure it's worth the tradeoff of limiting users to a fixed list of formats. As I said above, graphviz is pluggable, so there can be people using format plugins we don't even know about.

Dynamically generating an enum at runtime feels a bit fraught with peril, but in theory that would let us define a format list that's accurate for the current installation. In theory we could run a `dot -Thelp` command, parse the output, and there's our complete list of formats. I wonder if there's any way to make that work _and_ satisfy IntelliSense?

ferdnyc:
> Meaning, if someone has a custom-built graphviz with a `webp` plugin, or maybe a `blend` plugin for generating 3D models of their graphs, they can use it with pydot:

HAHAHA! Proving my point, graphviz _does_ have a built-in `webp` format now, which I guess was never added to our list:

```console
$ dot -Thelp
Format: "help" not recognized. Use one of: bmp canon cmap cmapx
 cmapx_np dot dot_json eps fig gd gd2 gif gtk gv ico imap imap_np
 ismap jpe jpeg jpg json json0 kitty kittyz mp pdf pic plain
 plain-ext png pov ps ps2 svg svgz tga tif tiff tk vrml vt vt-24bit
 wbmp webp x11 xdot xdot1.2 xdot1.4 xdot_json xlib
```

...There's actually a lot here, that's missing from `OUTPUT_FORMATS`.

ferdnyc:
> ...There's actually a lot here, that's missing from `OUTPUT_FORMATS`.

There are also a lot in `OUTPUT_FORMATS` that are no longer supported, at least by my installed version of `dot`. Mostly it's old crap that nobody wants to use anyway, like `hpgl` or `dia`, but it reinforces the point that even as an enum, our `OUTPUT_FORMATS` list **doesn't** determine the valid `format=` parameters that can be passed to `create()` or `write()`. The graphviz command does.

lkk7:
Interesting. Maybe during the first import of `pydot`, a command could be run and its result could be parsed to get available formats.
But that feels hacky, like scraping a website. What if there's an old or new version that will change the way it's displayed, etc...
Unless there is a command strictly made to get those formats in a clean way

PR #355, opened 2024-06-07
Quote graph/node name when setting it

tusharsadhwani:
This would avoid the graph breaking when setting the node name to for eg. something with spaces or slashes.

## Steps to reproduce:

```python
import pydot

graph = pydot.Graph()
graph.set_name("foo bar")
print(graph.to_string())
```

```console
$ python3 x.py
digraph foo bar {
}

$ python3 x.py | dot -Tpng
Error: <stdin>: syntax error in line 1 near 'bar'
```

### After fix:

```console
$ python3 x.py
digraph "foo bar" {
}
```

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -659,7 +659,7 @@ class Node(Common):
 
     def set_name(self, node_name):
         """Set the node's name."""
-        self.obj_dict["name"] = node_name
+        self.obj_dict["name"] = quote_if_necessary(node_name)
 
     def get_name(self):
         """Get the node's name."""
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -1042,7 +1042,7 @@ class Graph(Common):
 
     def set_name(self, graph_name):
         """Set the graph's name."""
-        self.obj_dict["name"] = graph_name
+        self.obj_dict["name"] = quote_if_necessary(graph_name)
 
     def get_name(self):
         """Get the graph's name."""


lkk7:
@tusharsadhwani Thanks for the PR.

There's already a quoting change going on in here: #339, and it's a much bigger change.
When that one is ready, we'll see how they are related and if they overlap.

Thank you!

ferdnyc:
@tusharsadhwani Like @lkk7 said, thanks for submitting this.

Quoting the name on _input_ has implications for name-matching in the object dictionaries, so I'd prefer to avoid that if possible.

For Node names, this definitely isn't necessary, because `Node.to_string()` will always call `quote_if_necessary()` on the `"name"` member when it's _output_:

https://github.com/pydot/pydot/blob/1ce0f34e0d23a69433d72f6896ddf2a2c3dbefbc/src/pydot/core.py#L682-L686
https://github.com/pydot/pydot/blob/1ce0f34e0d23a69433d72f6896ddf2a2c3dbefbc/src/pydot/core.py#L707-L710

What you've uncovered here is a bug, basically, in that Graph doesn't do the same, when it _should_:

https://github.com/pydot/pydot/blob/1ce0f34e0d23a69433d72f6896ddf2a2c3dbefbc/src/pydot/core.py#L1384-L1403

Instead, `graph` calls `quote_if_necessary()` on the name when it's _input_ (in `Graph.__init__`) and stores it that way.

The way I'd prefer to fix this is to make Graph work the same way Node already works: remove the call to `quote_if_necessary()` when the name is stored in `__init__()`, and instead call `quote_if_necessary()` when the `graph_name` is extracted during output.

The main reason I'd prefer to go that route — and the reason, I suspect, that `Node` already takes a different route from `Graph` — is that methods like `get_node()` will expect the name provided to search on to be the _same_ name that the Node was created with. So, changing the name behind-the-scenes as it's stored will screw up people's lookups. We really want to store the name exactly as it's provided by the user, and only parse/modify it when necessary on output.

(The current Graph functionality not only fails to properly output the names of graphs that have had `set_name()` called, but I suspect it fails to find subgraphs with quotable names using `get_subgraph()`, since the name to be looked up has to be the quoted form. That's why I prefer the way Node currently works.

A PR to fix this Graph-name-output bug would be welcome, though, and I don't think there's any reason to hold it pending #339; this is a clear and obvious bug, a fix can be incorporated into my #339 if necessary. (Although I don't think it will be, if it only touches Graph naming, because I hadn't planned to mess with that in #339.)

ferdnyc:
Huh. Well, I'd misremembered one thing: `Node` also calls `quote_if_necessary()` on the name, in its `__init__()`. It really shouldn't, because this is annoying:

```python3
>>> import pydot
>>> g = pydot.Graph("my graph")
>>> g.add_node(pydot.Node("my node"))
>>> g.get_node("my node")
[]
>>> g.get_node('"my node"')
[<pydot.core.Node object at 0x7faa47039a60>]
```

(Graph will do the same thing, currently.)

As long as `quote_if_necessary()` is called during _output_, IMHO there's no need to modify the name on _input_, since all it does is confuse lookups.

ferdnyc:
And like I said, calling `quote_if_necessary()` during _output_ means that you don't have to worry about quoting the name on input. This works fine, currently:
```python3
>>> import pydot
>>> g = pydot.Graph("my graph")
>>> n = pydot.Node("my node", color="red")
>>> n.set_name("my other node")
>>> g.add_node(n)
>>> g.get_node("my other node")
[<pydot.core.Node object at 0x7f72704e81a0>]
>>> print(g.to_string())
digraph "my graph" {
"my other node" [color=red];
}
```

...And that's the way I'd prefer things to work, even if the name is set by `Node.__init__()` or `Graph.__init__()`.

ferdnyc:
I also noticed that if you call `.set_name()` on a `Node` instance _after_ it's been added to a Graph, its name changes, but the reference under which it's stored in the Graph _doesn't_. So, that's a whole other bug, arguably:

```python3
>>> import pydot
>>> g = pydot.Graph("my graph")
>>> n = pydot.Node("my node", color="red")
>>> g.add_node(n)
>>> n.set_name("my other node")
>>> g.get_node("my other node")
[]
>>> g.get_node('"my node"')
>>> g.get_node('"my node"')[0].to_string()
'"my other node" [color=red];'
```

ferdnyc:
As it turns out, I likely _am_ going to have to address this as part of #339, because I'm adding new name-lookup logic to the parsing of `Edge` endpoints. (An edge endpoint string that exactly matches an existing name in the graph hierarchy will be presumed to represent that named entity, even if the name contains one or more colons.)

So, it'll be essential that names be stored without additional quoting, in order for the new lookup logic to properly identify edges that connect to those named entities.

tusharsadhwani:
Makes sense! I'm looking forward to it.

tusharsadhwani:
Can conifrm this is fixed in v3.0

PR #356, opened 2024-06-12
Move get_top_graph_type() to Common

ferdnyc:
A quick fix I noticed while working on #339.

By making `get_top_graph_type()` a method of all classes, and handling the lookup of parent graphs there, the code no longer needs to do laborious checks on the existence of a parent graph, and the existence of the parent graph's `get_top_graph_type()` method, just to be able to call it.

The method can now return `None`, if the current instance has no parent graph, but that's fine as we always compare the return value to some string. `None == "graph"` or `None == "digraph"` will always evaluate to `False`.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -569,6 +569,22 @@ class Common:
     def get_parent_graph(self):
         return self.obj_dict.get("parent_graph", None)
 
+    def get_top_graph_type(self):
+        """Find the topmost parent graph type for the current object."""
+        if isinstance(self, Graph):
+            parent = self
+        else:
+            parent = self.get_parent_graph()
+        if parent is None:
+            return None
+        while True:
+            parent_ = parent.get_parent_graph()
+            if parent_ == parent:
+                break
+            parent = parent_
+
+        return parent.obj_dict["type"]
+
     def set(self, name, value):
         """Set an attribute value by name.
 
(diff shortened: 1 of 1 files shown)


src/pydot/core.py
```diff
@@ -849,13 +865,8 @@ def to_string(self):
         else:
             edge = [src]
 
-        if (
-            self.get_parent_graph()
-            and self.get_parent_graph().get_top_graph_type()
-            and self.get_parent_graph().get_top_graph_type() == "digraph"
-        ):
+        if self.get_top_graph_type() == "digraph":
```
ferdnyc, line 868: This is the real win that comes from making it a method of every class.

src/pydot/core.py
```diff
@@ -569,6 +569,22 @@ def set_parent_graph(self, parent_graph):
...
         return self.obj_dict.get("parent_graph", None)
 
+    def get_top_graph_type(self):
+        """Find the topmost parent graph type for the current object."""
+        if isinstance(self, Graph):
+            parent = self
+        else:
+            parent = self.get_parent_graph()
+        if parent is None:
+            return None
```
ferdnyc, line 579: Now that I think about it, I can simplify this. We know that `Graph.get_parent_graph()` will always return the graph itself, so it'll just short-circuit the loop sooner.

```suggestion
    def get_top_graph_type(self):
        """Find the topmost parent graph type for the current object."""
        parent = self.get_parent_graph()
        if parent is None:
            return None
```

PR #359, opened 2024-06-16
Implement indentation for all to_string() methods

ferdnyc:
All `to_string()` methods now accept an `indent` keyword argument, which can be either a number or a string. `indentation=4` will indent each level of the output with four spaces. `indentation="  "` will indent each level by that many spaces. Other whitespace is also valid, e.g. `indent="\t"` for tab indents.

No indentation is performed by default, it will only be added if an initial non-zero/non-empty-string `indent` argument is given to whatever `to_string()` method is being called.

The methods also take additional arguments, meant for internal use:

  * `indent_level` is used to determine the number of times to insert the `indent` string before each line. The default for `Graph` is `0`, the default for `Node` and `Edge` is `1`. (This is so that e.g. `edge.to_string(indent=4)` will output an edge indented by 4 spaces). Each method increases the `indent_level` by 1 when calling other `to_string()` methods.

  * `Graph.to_string()` takes an `inline` argument, default `False`, which is used by `Edge.to_string()` when calling `Subgraph.to_string()` to output endpoints. It suppresses the indent on the first line, and the newline after the graph.

Also fixed some possible double-spacing:
  * In Graph when outputting anonymous subgraphs (`subgraph  {`)
  * In Edge before attributes (`A -- B  [...];`)

### TODO
- [ ] Could really use some tests
- [ ] Think of corner cases where it might break? (I've even tried it with `pydot.Dot` instances generated by `dot_parser.parse_dot_data()`, so it's successfully formatted edges with `frozendict` endpoints and the like.)


ferdnyc:
Example input/output:
```python
>>> from pydot import dot_parser
>>> graph_src="""graph G {
...     A [style=dotted, shape=box, color=red]
...     B [style=dashed, shape=square]
...     subgraph SG {
...         C [shape=box]
...         D [shape=circle]
...     }
...     A -- C [color=blue]
...     B -- { C D }
... }"""
>>> g = dot_parser.parse_dot_data(graph_src)[0]
>>> print(g.to_string())
graph G {
A [color=red, shape=box, style=dotted];
B [shape=square, style=dashed];
subgraph SG {
C [shape=box];
D [shape=circle];
}

A -- C [color=blue];
B -- subgraph {
C;
D;
};
}

>>> print(g.to_string(indent=4))
graph G {
    A [color=red, shape=box, style=dotted];
    B [shape=square, style=dashed];
    subgraph SG {
        C [shape=box];
        D [shape=circle];
    }

    A -- C [color=blue];
    B -- subgraph {
        C;
        D;
    };
}

>>> print(g.to_string(indent=" "))
graph G {
 A [color=red, shape=box, style=dotted];
 B [shape=square, style=dashed];
 subgraph SG {
  C [shape=box];
  D [shape=circle];
 }

 A -- C [color=blue];
 B -- subgraph {
  C;
  D;
 };
}

>>>
```

PR #360, opened 2024-06-16
Use ruff format --diff in CI

ferdnyc:
It'll still register as a failure, but at least whatever problems it has with the code will be visible.


ferdnyc:
Hmm, I should've injected some formatting errors somewhere, to demonstrate. Eh, I'll do that now, then drop the commit later.

lkk7:
Looks fine!

ferdnyc:
Dropping the extra commit!

Issue #358, opened 2024-06-16
Why does pydot sort attribute keys?


ferdnyc:
Currently, when `pydot` outputs a list of attributes, it uses `sorted()` to output them in alphabetical order.

But... why?

Since CPython 3.6, and since 3.7 it's part of the language definition, `dict` keys will always be returned in the order they were created. `**kwargs` is also guaranteed to preserve argument order.

Which means that, simply by **not** sorting the attributes, we _could_ output them in the same order they originally appeared, whether they were created by the `dot_parser` or by user code.

That would mean that doing this:
```python
import pydot
from pydot import dot_parser
graph_src = """graph G {
    A [style=dotted, shape=box, color=red]
    B [style=dashed, shape=square]
}"""
g = dot_parser.parse_dot_data(graph_src)[0]
print(g.to_string())
```
Would produce something a lot closer to the original source than this:
```dot
graph G {
A [color=red, shape=box, style=dotted];
B [shape=square, style=dashed];
}
```

lkk7:
That's a great point. Made a PR for this: https://github.com/pydot/pydot/pull/361, feel free to review

PR #361, opened 2024-06-16
Don't use `sorted()` on attributes in `to_string()` methods

lkk7:
Fix https://github.com/pydot/pydot/issues/358 (unnecessary sorting of attributes in `to_string()` methods)

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -700,7 +700,7 @@ class Node(Common):
 
         node_attr = []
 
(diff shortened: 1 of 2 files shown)


ferdnyc (approved):
Looks good to me! And I love that there were only two tests that were even affected by this. :grinning: (I wonder if that's a sign we have lots of tests with the attributes already in alphabetical order, or if it's a sign we need more multi-attribute tests?)

PR #363, opened 2024-06-17
Split attribute quoting from ID quoting, rewrite attribute formatting, store names unquoted

### LINKED ISSUE #85: Multiple styles not correctly rendered (quotes missing)

### LINKED ISSUE #111: graph node cannot be named "graph" or "node"
If a node in the graph named **`graph`** or **`node`**, `write_png()` function produces this error:

``` python
>>> import pydot
Couldn't import dot_parser, loading of dot files will not be possible.
>>> g = pydot.Dot()
>>> e = pydot.Edge('node', 'graph')
>>> g.add_edge(e)
>>> g.write_png('BUG.png')
Warning: /var/folders/_8/_99twmsd4zg5_09q7wm2y8jm0000gn/T/tmpdWTyp8: syntax error in line 2 near '->'

  File "<input>", line 1, in <module>
  File "/Library/Python/2.7/site-packages/pydot.py", line 1602, in <lambda>
    lambda path, f=frmt, prog=self.prog : self.write(path, format=f, prog=prog))
  File "/Library/Python/2.7/site-packages/pydot.py", line 1696, in write
    dot_fd.write(self.create(prog, format))
TypeError: must be string or buffer, not list
```

### LINKED ISSUE #180: get_node() and get_edge() does not match correctly if the original name was quoted

### LINKED ISSUE #246: Special chars Node naming making get_subgraph() access complicated
**SUMMARY**
Special Nodes names quoted, making get method complicated to use

**ISSUE TYPE**
help_wanted, details-needed

**COMPONENT NAME**
quote_if_necessary() 
get_subgraph() 

**ADDITIONAL INFORMATION**

Hello and thank you for maintaining this project. 

During my developments, I ran into an issue that bother me.

After creating a cluster using the Cluster class, 

```py
group = "SAP+"
main_graph.add_subgraph(pydot.Cluster(graph_name=group,
                                label=group))
```

The complex name attribute value  (**SAP+** here) is being quoted by quote_if_necessary() method.

My issue come after, at some point in my code, I try to access my created subgraph using get_subgraph() method. But when I'm using the base name I used to initialize it, I cannot find it back because it already have its quotes.

```py
subg = main_graph.get_subgraph("cluster_"+group)

(Pdb) subg
[]
(Pdb) main_graph.get_subgraphs()[0].get_name()
'"cluster_SAP+"'

(Pdb) main_graph.get_subgraph('"cluster_SAP+"')
[<pydot.Subgraph object at 0x7fd6597e7700>]
```
 
To fix this, I have to format specifically the name and add some exceptions:

```py
subg = main_graph.get_subgraph(group)
        # Case it's a special chars name
        if not subg:
            main_graph.get_subgraph('"{0}"'.format(group))
        # In case it's cluster (most of the cases ^^)
        if not subg:
            subg = main_graph.get_subgraph("cluster_"+group)
            # Second case when quoted name (SAP+ for ex...)
            if not subg:
                subg = main_graph.get_subgraph('"{0}{1}"'.format("cluster_", group))
```

I'm not sure we can say it's a bug but it's very inconvenient for accessing easily to the subgraphs.

Is there a better way to access this subgraph ? Should the method be enhanced/fixed to support this specific case ? 

The real root cause looks like a mix-up of language to me.

Regards

### LINKED ISSUE #258: String attributes containing colon are not quoted

### LINKED ISSUE #282: Graph object cannot have name "graph" as it breaks GraphViz
Naming Graph "graph", eg. `graph = pydot.Dot("graph", graph_type="graph")`
leads to following Dot string:
```
graph graph {
1 [label=ROOT];
2 [label=S];
2 -- 1;
3 [label=NP];
3 -- 2;
4 [label=VP];
4 -- 2;
5 [label=DT];
5 -- 3;
6 [label=VBZ];
6 -- 4;
7 [label=NP];
7 -- 4;
8 [label=This];
8 -- 5;
9 [label=is];
9 -- 6;
10 [label=DT];
10 -- 7;
11 [label=NN];
11 -- 7;
12 [label=a];
12 -- 10;
13 [label=test];
13 -- 11;
}
```
and following GraphViz error on any interaction:
```
"dot" with args ['-Tpng', '/tmp/tmpa165fw3n'] returned code: 1

stdout, stderr:
 b''
b"Error: /tmp/tmpa165fw3n: syntax error in line 1 near 'graph'\n"
```

### LINKED ISSUE #358: Why does pydot sort attribute keys?

ferdnyc:
As the scope of #339 has creeped rather a lot and threatened to push it into a state of perpetual limbo, I've started splitting out some of the changes into more focused PRs. This is another one of those, though the rather far-ranging nature might make it seem otherwise. This PR affects:

#### Attribute formatting
* Separate ID and attribute quoting, the same way #339 started out doing
* Fix quoting for attribute values
* Completely _rewrite_ the attribute **formatting** process, moving the meat of the work into new, centralized `Common` methods, which all classes' `to_string()` methods now call to get their formatted attributes
* Remove the attribute sorting, stomping on @lkk7's #361 (sorry! I did cherry-pick your test fixes into this PR, though.)

#### Name storage & formatting
* Remove all quoting of ID strings on _input_, switching to instead always performing quoting on _output_, so that strings are stored internally the same way the user inputs them, making lookups via `.get_*()` methods less confusing/annoying.
* Take advantage of the fact that we know when we're outputting a name vs. some other syntax, and quote graphviz keywords used as names to avoid syntax errors in the generated source. Make exception for the default-values node names.

Closes #361 
Closes #355
Closes #181
Fixes #358 
Fixes #282 
Fixes #258 
Fixes #246 
Fixes #180 
Fixes #111 
Fixes #85

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -291,14 +291,11 @@ class frozendict(dict):
 def make_quoted(s):
     """Transform a string into a quoted string, escaping specials."""
     replace = {
-        '"': r"\"",
-        "\n": r"\n",
-        "\r": r"\r",
+        ord('"'): r"\"",
+        ord("\n"): r"\n",
+        ord("\r"): r"\r",
     }
-    s.translate(replace)
-    for a, b in replace.items():
-        s = s.replace(a, b)
-    return rf'"{s}"'
+    return rf'"{s.translate(replace)}"'
 
 
 dot_keywords = ["graph", "subgraph", "digraph", "node", "edge", "strict"]
@@ -678,7 +675,8 @@ class Common:
     def attrs_string(self, prefix=""):
         """Format the current attributes list for output.
 
-        The `prefix` string will be prepended iff text is generated."""
+        The `prefix` string will be prepended if and only if some
+        output is generated."""
         attrs = self.formatted_attr_list()
         if not attrs:
             return ""
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -288,19 +288,49 @@ class frozendict(dict):
         return f"frozendict({dict_repr})"
 
 
+def make_quoted(s):
+    """Transform a string into a quoted string, escaping specials."""
+    replace = {
+        '"': r"\"",
+        "\n": r"\n",
+        "\r": r"\r",
+    }
+    s.translate(replace)
+    for a, b in replace.items():
+        s = s.replace(a, b)
+    return rf'"{s}"'
+
+
 dot_keywords = ["graph", "subgraph", "digraph", "node", "edge", "strict"]
 
-id_re_alpha_nums = re.compile(r"^[_a-zA-Z][a-zA-Z0-9_\.]*$", re.UNICODE)
+re_all_numeric = re.compile(r"^[0-9\.]+$")
+re_dbl_quoted = re.compile(r'^".*"$', re.S)
+re_html = re.compile(r"^<.*>$", re.S)
+
+id_re_alpha_nums = re.compile(r"^[_a-zA-Z][a-zA-Z0-9_\.]*$")
 id_re_alpha_nums_with_ports = re.compile(
-    r'^[_a-zA-Z][a-zA-Z0-9_\.:"]*[a-zA-Z0-9_\."]+$', re.UNICODE
+    r'^[_a-zA-Z][a-zA-Z0-9_\.:"]*[a-zA-Z0-9_\."]+$'
 )
-id_re_num = re.compile(r"^[0-9\.]+$", re.UNICODE)
-id_re_with_port = re.compile("^([^:]*):([^:]*)$", re.UNICODE)
-id_re_dbl_quoted = re.compile('^".*"$', re.S | re.UNICODE)
-id_re_html = re.compile("^<.*>$", re.S | re.UNICODE)
+id_re_with_port = re.compile(r"^([^:]*):([^:]*)$")
 
 
-def needs_quotes(s):
+def any_needs_quotes(s):
+    """Determine if a string needs to be quoted.
+
+    Returns True, False, or None if the result is indeterminate.
+    """
+    has_high_chars = any(ord(c) > 0x7F or ord(c) == 0 for c in s)
+    if has_high_chars and not re_dbl_quoted.match(s) and not re_html.match(s):
+        return True
+
+    for test_re in [re_all_numeric, re_dbl_quoted, re_html]:
+        if test_re.match(s):
+            return False
+
+    return None
+
+
+def id_needs_quotes(s):
     """Checks whether a string is a dot language ID.
 
     It will check whether the string is solely composed
@@ -319,19 +349,12 @@ def needs_quotes(s):
     if s in dot_keywords:
         return False
 
-    has_high_chars = any(ord(c) > 0x7F or ord(c) == 0 for c in s)
-    if (
-        has_high_chars
-        and not id_re_dbl_quoted.match(s)
-        and not id_re_html.match(s)
-    ):
-        return True
+    any_result = any_needs_quotes(s)
+    if any_result is not None:
+        return any_result
 
     for test_re in [
         id_re_alpha_nums,
-        id_re_num,
-        id_re_dbl_quoted,
-        id_re_html,
         id_re_alpha_nums_with_ports,
     ]:
         if test_re.match(s):
@@ -339,12 +362,12 @@ def needs_quotes(s):
 
     m = id_re_with_port.match(s)
     if m:
-        return needs_quotes(m.group(1)) or needs_quotes(m.group(2))
+        return id_needs_quotes(m.group(1)) or id_needs_quotes(m.group(2))
 
     return True
 
 
-def quote_if_necessary(s):
+def quote_id_if_necessary(s):
     """Enclose attribute value in quotes, if needed."""
     if isinstance(s, bool):
         if s is True:
@@ -357,18 +380,27 @@ def quote_if_necessary(s):
     if not s:
         return s
 
-    if needs_quotes(s):
-        replace = {
-            '"': r"\"",
-            "\n": r"\n",
-            "\r": r"\r",
-        }
-        for a, b in replace.items():
-            s = s.replace(a, b)
-
-        return '"' + s + '"'
+    if id_needs_quotes(s):
+        return make_quoted(s)
 
     return s
+
+
+def quote_attr_if_necessary(s):
+    if isinstance(s, bool):
+        return str(s)
+
+    if not isinstance(s, str):
+        return s
+
+    if s in dot_keywords:
+        return make_quoted(s)
+
+    any_result = any_needs_quotes(s)
+    if any_result is not None and not any_result:
+        return s
+
+    return make_quoted(s)
 
 
 def graph_from_dot_data(s):
@@ -664,7 +696,7 @@ class Node(Common):
             if isinstance(name, int):
                 name = str(name)
 
-            self.obj_dict["name"] = quote_if_necessary(name)
+            self.obj_dict["name"] = quote_id_if_necessary(name)
             self.obj_dict["port"] = port
 
     def __str__(self):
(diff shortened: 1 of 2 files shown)


ferdnyc:
(Actually I think #85 was already fixed by #320, but it never got closed so might as well close it here.)

src/pydot/core.py
```diff
@@ -288,19 +288,52 @@ def __repr__(self):
...
 
 
+def make_quoted(s):
+    """Transform a string into a quoted string, escaping specials."""
+    replace = {
+        '"': r"\"",
+        "\n": r"\n",
+        "\r": r"\r",
+    }
+    s.translate(replace)
```
lkk7, line 298: Looks like the string is translated twice here, once through `s.translate()` (but its result isn't assigned to anything), and once through the `for` loop.

src/pydot/core.py
```diff
@@ -618,6 +659,31 @@ def get_sequence(self):
...
+        """Return a list of the class's attributes as formatted strings."""
+        return [
+            self._format_attr(k, v)
+            for k, v in self.obj_dict["attributes"].items()
+        ]
+
+    def attrs_string(self, prefix=""):
+        """Format the current attributes list for output.
+
+        The `prefix` string will be prepended iff text is generated."""
```
lkk7, line 681:
```suggestion
        The `prefix` string will be prepended if text is generated."""
```

lkk7 (commented):
Great change, thanks.

I believe that the massive test suite with all the dot files should cover these changes safely?

src/pydot/core.py
ferdnyc: Heh. I actually did mean "iff" here, a shorthand for "if and only if". But that's probably a bit jargony, so I should write it out.

src/pydot/core.py
ferdnyc: You're right, I don't know what I was thinking when I wrote this.

There's no need for the loop at all, it should just `return rf'"{s.translate(replace)}"'` at the end.

ferdnyc:
The tests _probably_ do cover all of this, but I'd feel better adding some extra tests specifically for the rewritten functionality.

I started writing some, then stumbled on the fact that it isn't always easy to compare values in the test suite. (On the plus side, _trying_ to compare Node values was how I stumbled on the broken parenting behavior that led to #365.)

src/pydot/core.py
ferdnyc: Additionally, I appear to have completely forgotten that `str.translate()` takes a mapping of _ordinals_ to ordinals-or-strings. So the correct mapping is:

```python
replace = {
    ord('"'): r"\"",
    ord("\n"): r"\n",
    ord("\r"): r"\r",
}
```

Fixed now.

src/pydot/core.py
ferdnyc: Fixed to be less fake-clever.

ferdnyc:
@lkk7 

OK, I have just now added sufficient new tests targeted specifically at the functionality in this PR, that I think I can declare it working and verified. (I also fixed the items brought up in review.)

I hit on a method of testing that the `get_{node,edge,etc.}()` methods were returning the expected objects (even though they won't be the _same_ Python objects, because they're recreated each time a value is returned) by comparing the `id()` value of the `obj_dict` _inside_ each object.

If, for example, `Graph.get_node()` returns a particular `Node` object, it won't be the same `Node` object that was originally passed to `Graph.add_node()`, but it **will**  contain the same `obj_dict` that was originally passed in, because those are all stored by reference and don't get copied internally. So, code like this in the tests works, where direct comparison of the objects would not:

```python
g = pydot.Graph()
a = pydot.Node("my node")
g.add_node(a)
a_out = g.get_node("my node")[0]
self.assertEqual(id(a_out.obj_dict), id(a.obj_dict))
```

I may even define `__eq__` dunder methods for `Node`, `Edge`, etc. that implement that check, so that we can use `some_node == some_returned_node` or `self.assertEqual(some_node, some_returned_node)` and get the correct results. (It'd make the tests I just wrote a _lot_ cleaner, for one.) But that can be a followup PR.

ferdnyc:
(If I'm not mistaken all I've done there is a poor-man's reimplementation of `is`, since `a is b` is effectively a concise way of saying `id(a) == id(b)`. unittest _does_ have an `assertIs()`, so the test could theoretically also be written:

```python
g = pydot.Graph()
a = pydot.Node("my node")
g.add_node(a)
a_out = g.get_node("my node")[0]
self.assertIs(a_out.obj_dict, a.obj_dict)
```

lkk7 (approved):
Great PR, thank you!
I'll check and close all these issues (where applicable) this weekend.

tusharsadhwani:
Amazing work!

lkk7:
@ferdnyc Well, I shouldn't close the issues until the fixes are properly released.
I think now is a good time for a release, so I'll prepare a changelog

ferdnyc:
@lkk7 

Oh... Oops. Sorry, they already closed. The PR was set to auto-close all of them.

>  I shouldn't close the issues until the fixes are properly released.

I mean, I guess there are two schools of thought on that. As soon as the fix is in the repo, the issue is solved. Anyone who wants to run with the `master` branch can always install the development code with:

```bash
pip install git+https://github.com/pydot/pydot.git
```

(Hmmm... we should probably add that to the README.)

ferdnyc:
(Ugh, that reminds me. As much of a pain in the ass as it can be, it's probably a mitzvah if we take the time to rename the `master` branch to `main`.)

lkk7:
Oh, that's fine. The release will be coming soon anyway. 

I'll add that note to README.md and I'll change the branch name.

PR #367, opened 2024-06-22
authors += @ferdnyc, @lkk7; maintainers += @lkk7

ferdnyc:
The package metadata just seemed objectively out of sync with recent reality. Figure it's past time we {claim some credit | accept our share of blame} for this nonsense.

@lkk7: I just copied<sup>1</sup> the name and email fields from your commits verbatim, feel free to self-apply suggestions to adjust them however you prefer.

And feel free to reject outright, if you're in any way hesitant about any part of this change.

<sup>1</sup>: (Technically I _retyped_ your email address, actually, so please double-check my spelling.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,9 +14,12 @@ dependencies = [
 authors = [
   {name = "Ero Carrera", email = "ero.carrera@gmail.com"},
   {name = "Peter Nowee", email = "peter@peternowee.com"},
+  {name = "Łukasz", email = "lukaszlapinski7@gmail.com"},
+  {name = "FeRD (Frank Dana)", email = "ferdnyc@gmail.com"},
 ]
 maintainers = [
   {name = "Peter Nowee", email = "peter@peternowee.com"},
+  {name = "Łukasz", email = "lukaszlapinski7@gmail.com"},
 ]
 keywords = ["graphviz", "dot", "graphs", "visualization"]
 classifiers = [


lkk7 (approved):

lkk7:
Good idea. 
On a similar note, I was also thinking about preparing the changelog for the upcoming version. If all the changes get into it (together with all the upcoming ones fixing quoting), then it'll probably have to be 3.0 right after a 2.0 🥲

ferdnyc:
@lkk7 
> If all the changes get into it (together with all the upcoming ones fixing quoting), then it'll probably have to be 3.0 right after a 2.0 🥲

I was actually going to suggest the same thing, I think by the time the quoting-behavior fixes and the other changes I'm looking to make are all merged, it'll be impossible for there **not** to be some breaking changes, so it's best to do a major version bump.

And then we have the leeway to do things like outright **remove** the support for port arguments on `Node` names. On the simple grounds that the code's never worked right, and even if it were fixed those values are literally meaningless to graphviz. So there's really no point in even bothering with a deprecation period, we already know nobody's using that feature for anything because there's nothing to use it _for_.

(Port/compass values on `Edge` endpoints would stay, of course, and in fact I believe I should have the support significantly improved and cleaned up within the next week or two. But the `Node` support is completely useless code.)

ferdnyc:
Which reminds me, there's **also** some code I noticed in the parser, dating all the way back to the code's very first commit, that implements support for syntax that AFAICT is completely invalid. For some reason, there's support in the parser for consuming code that graphviz itself would reject with a syntax error. (I don't know if maybe it's _previously-supported_ syntax that was removed so long ago there isn't even any record of it anymore, or what.) So, that's on my chopping block too.

It's this bit (in the port parsing, naturally):

https://github.com/pydot/pydot/blob/b97c4df88f3989b65ccb6ccf2b41b033f3b3d235/src/pydot/dot_parser.py#L413-L423

I mean, WTAF?

A `port_angle` that takes the form `@<id>`?
A second form of `port_location` specified as `:(<id>,<id>)`?
Combined `port` value formats like `:(<id>,<id>)@<id>` or `@<id>:(<id>,<id>)`?

_None_ of that is accepted by `dot`. I tried! It's all syntax errors. So, that's a nice little chunk of (presumably never-accessed) nonsense code we can drop-kick from the parser.

ferdnyc:
> So, that's on my chopping block too.

(I'll likely inquire about it over @ the graphviz forum first, because... well, because I'm morbidly curious, basically.

Maybe one of the old guard there will have some insights into what any of that syntax could possibly mean, whether anything actually does accept it, if there was _ever_ anything that accepted it, or was whoever wrote those bits of the parser just inventing new grammar on the fly in the hopes of making `@` happen.)

Issue #368, opened 2024-06-22
Support for invalid syntax in dot_parser?


ferdnyc:
Dating all the way back to the very first `dot_parser.py` commit, there's a section of code that implements support for syntax that AFAICT is completely invalid.

I have no idea why the parser accepts most of the syntax defined here, because I can't find any sign that it could ever be valid. Not only is none of it documented in the official graphviz grammar, but I've attempted to parse it with `dot` and it's all rejected as a syntax error.

(Maybe it's _previously-supported_ syntax that was removed so long ago there isn't even any record of it anymore... I don't know.) But unless we can find some scenario where any of this code would actually be valid, it seems safe to remove it from the parser since it'll never be encountered in valid graphviz source.

It's this bit (in the port parsing, naturally):

https://github.com/pydot/pydot/blob/b97c4df88f3989b65ccb6ccf2b41b033f3b3d235/src/pydot/dot_parser.py#L413-L423

As far as I can tell, graphviz never supports and will never produce/accept:
* A `port_angle` that takes the form `@<id>`
* A second form of `port_location` specified as `:(<id>,<id>)`
* Combined `port` value formats like `:(<id>,<id>)@<id>` or `@<id>:(<id>,<id>)`

_Edited from a comment originally posted by @ferdnyc in https://github.com/pydot/pydot/issues/367#issuecomment-2183953447_

ferdnyc:
Copying my update from https://github.com/pydot/pydot/pull/377#issuecomment-2227099171

I've been continuing to have conversations with the Graphviz team about this. The [latest development](https://forum.graphviz.org/t/strange-graphviz-syntax-in-pydot-parser-valid-useful-anywhere/2301/3) is that Steve Roush actually _found_ all of this syntax documented in the grammar published in the "Drawing graphs with _dot_" paper, AKA the [old PDF documentation](https://graphviz.org/pdf/dotguide.pdf#page=34).

That this was as much of a surprise to most of the Graphviz team, as it was to me, definitely makes me feel better.

Stephen North (who's one of the three authors of the aforementioned PDF, so he should know) responded,

> We meant these grammar hooks for features that for the most part we never got around to developing. Sorry.

So, it appears this isn't so much _old_ syntax, as never-implemented syntax. Current versions of `dot` definitely do not support it, they'll abort with a syntax error if they encounter either an `@` sign, or a `:(`, when parsing an ID.

So, yes, it still appears that we would be correct to remove this syntax, and do the same.

PR #377, opened 2024-07-13
Remove support for old invalid syntax

lkk7:
If:
- `port_angle` is not supported,
- `Group(colon + lparen + ID + comma + ID + rparen)` is not supported,

Then the resulting grammar code can be simplified.

### DIFF
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -372,12 +372,9 @@ def graph_definition():
         rbrace = Literal("}")
         lbrack = Literal("[")
         rbrack = Literal("]")
-        lparen = Literal("(")
-        rparen = Literal(")")
         equals = Literal("=")
         comma = Literal(",")
         semi = Literal(";")
-        at = Literal("@")
         minus = Literal("-")
 
         # keywords
@@ -410,17 +407,7 @@ def graph_definition():
 
         righthand_id = (float_number | ID).setName("righthand_id")
 
-        port_angle = (at + ID).setName("port_angle")
-
-        port_location = (
-            OneOrMore(Group(colon + ID))
-            | Group(colon + lparen + ID + comma + ID + rparen)
-        ).setName("port_location")
-
-        port = (
-            Group(port_location + Optional(port_angle))
-            | Group(port_angle + Optional(port_location))
-        ).setName("port")
+        port = Group(OneOrMore(Group(colon + ID))).setName("port")
 
         node_id = ID + Optional(port)
         a_list = OneOrMore(


lkk7:
@ferdnyc Related to what you found here: https://github.com/pydot/pydot/issues/368

ferdnyc:
@lkk7 

I've been continuing to have conversations with the Graphviz team about this. The [latest development](https://forum.graphviz.org/t/strange-graphviz-syntax-in-pydot-parser-valid-useful-anywhere/2301/3) is that Steve Roush actually _found_ all of this syntax documented in the grammar published in the "Drawing graphs with _dot_" paper, AKA the [old PDF documentation](https://graphviz.org/pdf/dotguide.pdf#page=34).

That this was as much of a surprise to most of the Graphviz team, as it was to me, definitely makes me feel better.

Steven North (who's one of the three authors of the aforementioned PDF, so he should know) responded,

> We meant these grammar hooks for features that for the most part we never got around to developing. Sorry.

So, it appears this isn't so much _old_ syntax, as never-implemented syntax. Current versions of `dot` definitely do not support it, they'll abort with a syntax error if they encounter either an `@` sign, or a `:(`, when parsing an ID.

So, yes, it still appears that we would be correct to remove this syntax, and do the same.

src/pydot/dot_parser.py
```diff
@@ -372,12 +372,9 @@ def graph_definition():
         rbrace = Literal("}")
         lbrack = Literal("[")
         rbrack = Literal("]")
-        lparen = Literal("(")
-        rparen = Literal(")")
         equals = Literal("=")
         comma = Literal(",")
         semi = Literal(";")
-        at = Literal("@")
         minus = Literal("-")
```
ferdnyc, line 378: This is tangential, but all of these `Literal()` aliases actually really annoy me. Unlike the `CaselessKeyword()`s that precede them, which _are_ needed for case-insensitive parsing, explicit `Literal()`s are **totally** unnecessary.

Just using a bare string value in a pyparsing expression **implies** `Literal()`. So, `"="` is equivalent to `Literal("=")` in every way except that it isn't a `ParserElement`, so you can't call methods on it like `"=".suppress()`.

But that's not necessary anyway, because `Suppress()` exists, and like every other pyparsing grammar class is a subclass of `ParserElement`. So IMHO, the grammar is more readable when a rule like this,

```python
graph_stmt = Group(
    lbrace.suppress()
    + Optional(stmt_list)
    + rbrace.suppress()
    + Optional(semi.suppress())
).setName("graph_stmt")
```
is instead written as,

```python
graph_stmt = Group(
    Suppress("{") + Optional(stmt_list) + Suppress("}") + Optional(";").suppress()
).setName("graph_stmt")
```
or even (not sure if this would work):

```python
    Suppress("{") + Optional(stmt_list) + Suppress("}" + Optional(";"))
```

src/pydot/dot_parser.py
```diff
@@ -410,17 +407,7 @@ def graph_definition():
...
-        port_location = (
-            OneOrMore(Group(colon + ID))
-            | Group(colon + lparen + ID + comma + ID + rparen)
-        ).setName("port_location")
-
-        port = (
-            Group(port_location + Optional(port_angle))
-            | Group(port_angle + Optional(port_location))
-        ).setName("port")
+        port = Group(OneOrMore(Group(colon + ID))).setName("port")
```
ferdnyc, line 410: I wonder if this should be a `OneOrMore`, or if it should explicitly limit the number of colons to 2? IOW,

```python
port = (
    Group(":" + ID + ":" + ID)
    | Group(":" + ID)
).set_name("port")
```
(In that order because `|` alternation takes the first branch that matches.)

I'd like to `Suppress()` the leading colon, as well, but our current weird logic keeps it as part of the `obj_dict["port"]`. So I think it has to be captured here as well, until/unless we fix that.

ferdnyc (approved):
Some commentary / alternatives, but nothing that affects the code as-is. LGTM!

src/pydot/dot_parser.py
ferdnyc: Should've known that'd be too wide for GitHub's narrow comment rendering.

lkk7: Noted for the future.

src/pydot/dot_parser.py
lkk7: Noted for the future. 

I even changed it but it leads to test failures. Because of some logic in the `do_node_ports`, the `Group` wrapping has to be kept:
```python
port = Group(
    Group(":" + ID + ":" + ID)
    | Group(":" + ID)
).setName("port")
```

But even with that, a single test (`ports.dot`) is still failing. So that will need a logic change.

ferdnyc: Hmm, odd. I'll try to take a look at that, at some point.

As another mid-term thing, the parsing code can actually be improved a _lot_ if we switch to using results names.

Right now, all of the parse actions walk the token list picking out potentially interesting things to process. But there's a much easier way.

Say I have a `node_stmt` defined like this:
```python
node_stmt = (
    node_id.set_results_name("node_id")
    + Optional(attr_block).set_results_name("node_attrs")
    + Optional(';').suppress()
)
```
(Where `node_id` and `attr_block` are defined appropriately.)

Then, when I parse a node line, the relevant items will automatically be identified with the results names assigned, and can be accessed that way directly. (`node_stmt` in this example has no parse actions set on it, initially.)

```python
>>> from pprint import pp
>>> res = node_stmt.parse_string(
... "a [color=red, shape=box];")
>>> pp(res.as_dict())
{'node_id': ['a'],
 'node_attrs': [['color', 'red'], ['shape', 'box']]}
>>> res['node_id']
ParseResults(['a'], {})
```

...And they can be accessed that way from _inside_ the parse actions, as well.

```python
>>> def demo(toks):
...    _id = toks.get('node_id')
...    attrs = toks.get('node_attrs')
...    print(f"{_id}\n{attrs}")
...
>>> node_stmt.set_parse_action(demo)
node_stmt
>>> res = node_stmt.parse_string(
... "a [color=red, shape=box];")
['a']
[['color', 'red'], ['shape', 'box']]
>>> 
```

Used correctly, it's a much more deterministic way of consuming the parse results than just walking the raw token list. By doing that, the parse action is effectively re-parsing the input. A lot of pyparsing's power is going to waste, since it's just acting as an input validator and dumb tokenizer.

ferdnyc: I figured it out. Because of the assumptions about token hierarchy in `push_edge_stmt()` and `do_node_ports()`, it has to be:

```python
port = (
    Group(Group(colon + ID) + Group(colon + ID))
    | Group(Group(colon + ID))
).setName("port")
```

...But that does work. I'll put together a PR.

PR #378, opened 2024-07-13
Prepare changelog for 3.0.0

lkk7:


lkk7:
@ferdnyc as you've made the most important change for this upcoming release, feel free to suggest a more verbose version of it because you'll do it better for sure.

ChangeLog
```diff
@@ -4,14 +4,19 @@ Pydot versions since 1.4.2 adhere to [PEP 440-style semantic versioning]
...
+3.0.0 (unreleased)
 ------------------
 
 Added:
+- Breaking change: support for Python 3.7 dropped totally. (#371)
+  It was already dropped, but now it's official.
+- Breaking change (but for nobody): Old unknown syntax removed (#377).
+  A "port angle" that takes the form `@<id>`
+  and an alternative form of a port `:(<id>,<id>)` deleted.
+- Quoting for attribute values fixed (#363).
```
ferdnyc, line 16:
```suggestion
- Breaking change (but for nobody): Invalid syntax removed (#377).
  Removed syntax rules that were never implemented in graphviz's own parser.
- Attribute sorting removed (#361).
  Pydot will preserve the original order of attributes as defined.
- Quoting for attribute values fixed (#320).
  Attribute values containing comma-separated strings will now be quoted correctly.
- Internal storage and lookup of identifiers (names) improved (#363).
  Pydot now always stores values as they were originally input, and only performs
  quoting on output.
  Previously, when a graph element was created with a name that contained a space
  or otherwise required quoting, the quotes would be added immediately, making it
  more difficult to retrieve the element using the `.get_*()` lookup methods.
  If a name requiring quotes was later set using .set_name(), no quotes would be
  added, causing the graph definition to become invalid.
  (Thanks to @tusharsadhwani for initially pointing out the name-quoting issues.)
- Keywords can now be used as names or attribute values (#363).
  Graphviz keywords like "graph" or "subgraph" will now be properly quoted
  when used as attribute values or as names, where appropriate.
```

ChangeLog
```diff
@@ -4,14 +4,19 @@ Pydot versions since 1.4.2 adhere to [PEP 440-style semantic versioning]
 (https://www.python.org/dev/peps/pep-0440/#semantic-versioning).
 
 
-2.0.1 (unreleased)
+3.0.0 (unreleased)
 ------------------
 
 Added:
+- Breaking change: support for Python 3.7 dropped totally. (#371)
+  It was already dropped, but now it's official.
```
ferdnyc, line 12: Oh, and just so it matches the others...

```suggestion
- Breaking change: support for Python 3.7 dropped totally (#371).
  It was already dropped, but now it's official.
```

ChangeLog
ferdnyc: (IMHO the specifics of the removed syntax are unnecessary detail. Nobody could use it _before_, and they can't use it now. But, your call.)

ChangeLog
```diff
@@ -4,14 +4,19 @@ Pydot versions since 1.4.2 adhere to [PEP 440-style semantic versioning]
...
+- Breaking change (but for nobody): Old unknown syntax removed (#377).
+  A "port angle" that takes the form `@<id>`
+  and an alternative form of a port `:(<id>,<id>)` deleted.
```
ferdnyc, line 15:
```suggestion
- Breaking change (but for nobody): Invalid syntax removed (#377).
  Removed syntax rules that were never implemented in graphviz's own parser.
```

(IMHO the specifics of the removed syntax are unnecessary detail. Nobody could use it before, and they can't use it now. But, your call.)

ChangeLog
ferdnyc, line 16: The remainder...

```suggestion
- Internal storage and lookup of identifiers (names) improved (#363).
  Pydot now always stores values as they were originally input, and only performs
  quoting on output.
  Previously, when a graph element was created with a name that contained a space
  or otherwise required quoting, the quotes would be added immediately, making it
  more difficult to retrieve the element using the `.get_*()` lookup methods.
  If a name requiring quotes was later set using `.set_name()`, no quotes would be
  added, causing the graph definition to become invalid.
  (Thanks to @tusharsadhwani for initially pointing out the name-quoting issues.)
- Keywords can now be used as names or attribute values (#363).
  Graphviz keywords like "graph" or "subgraph" will now be properly quoted
  when used as attribute values or as names, where appropriate.
- Attribute sorting removed (#361).
  Pydot will preserve the original order of attributes as defined.
- Quoting for attribute values fixed (#320).
  Attribute values containing comma-separated strings will now be quoted correctly.
```

ferdnyc (commented):
Submitted two versions of this, I'll delete the previous one.

ferdnyc:
Part of me wants to include #365 on the list, just in case it affects anyone, but it's probably too internal to give a concise, coherent explanation in a changelog entry. (I'm already pushing it with #363.)

PR #382, opened 2024-07-15
Changelog: Separate into categories

ferdnyc:
If I haven't missed the 3.0.0 release, I just realized that the Changelog listed everything as "Added", which isn't entirely accurate.


lkk7:
Woah, that's 100% my fault for losing my vigilance after committing your suggestions.
Good thing it's just a cosmetic thing.

PR #384, opened 2024-07-16
Fix edge quoting (broken in 3.0.0)

### LINKED ISSUE #383: Broken to_string() for nodes that contain '^' in 3.0.0
In case of node name collect `^`, edge saves node names without quotes.

```python
import pydot
graph = pydot.Dot("")
graph.add_node(pydot.Node("Node^A"))
graph.add_node(pydot.Node("Node^B"))
graph.add_edge(pydot.Edge("Node^A", "Node^B"))
data = graph.to_string()
print(data)
n_graph = pydot.graph_from_dot_data(data)
```

### pydot==3.0.0 

```
digraph  {
"Node^A";
"Node^B";
Node^A -> Node^B;
}

Node^A -> Node^B;
    ^
Expected graph_stmt, found '^'  (at char 35), (line:4, col:5)
None
```
### pydot==2.0.0

```
digraph  {
"Node^A";
"Node^B";
"Node^A" -> "Node^B";
}

[<pydot.core.Dot object at 0x7618654bf670>]
```

ferdnyc:
I missed a crucial call to `quote_id_if_necessary()` in the `to_string()` path for `pydot.Edge`, so it's broken in 3.0.0 and leaves some values unquoted that need to be quoted.

This PR adds a fix and a test. I'd suggest an immediate bugfix release of 3.0.1.

Fixes #383

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,7 +9,10 @@ Pydot versions since 1.4.2 adhere to [PEP 440-style semantic versioning]
 3.0.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+This is a bugfix release to correct a critical error introduced in 3.0.0.
+
+Changed:
+- Fix quoting in `Edge.to_string()`. (#384)
 
 
 3.0.0 (2024-07-15)
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -903,7 +903,7 @@ class Edge(Common):
 
             return node
 
-        return node_str
+        return quote_id_if_necessary(node_str)
 
     def to_string(self):
         """Return string representation of edge in DOT language."""
--- a/test/test_pydot.py
+++ b/test/test_pydot.py
@@ -397,6 +397,23 @@ class TestGraphAPI(PydotTestCase):
                 """),
         )
 
+    def test_edge_quoting(self):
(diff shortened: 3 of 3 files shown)


ferdnyc:
(Somehow NONE of our massive test suite exercised that code path, which is pretty amazing. I think it's because most of the tests use the pydot parser to _create_ the test data, and when `dot_parser` reads a quoted value in a graph, it'll include the quotes in the name. So, none of the tests that work from `.dot` files actually test the automatic quoting logic, as that can only be triggered using the Python API.)

That probably means we should write a **bunch** more tests that _don't_ rely on the parser to create the test data.

lkk7 (approved):

dulinriley:
I think a similar fix to this is needed for issue https://github.com/pydot/pydot/issues/418

ferdnyc:
@dulinriley Oh, no, this was a fix for my quoting rewrite in #363, where I just left out a quoting call.

The bad quoting _logic_ that led to #418 happened even earlier — way back in #320, in fact, when I let a fix for attribute quoting spill over into the ID quoting and (unknowingly) broke it.

(In #363 I ended up _splitting_ the attribute and ID quoting logic entirely, since they turned out to have more differences than similarities... but the damage was already done.)

PR #391, opened 2024-07-18
Update MANIFEST.in for REUSE config change

ferdnyc:
The REUSE config was relocated within the repo. Update the manifest accordingly.


lkk7 (approved):

ferdnyc:
I didn't change the `pyproject.toml` here, we should still figure out what we want to do with that.

PR #392, opened 2024-07-19
Ensure license files end up in wheel

### LINKED ISSUE #390: LICENSE file missing in pydot 3 wheels
In pydot 2, a `LICENSE` file was included in the `dist-info` folder of the wheel. In pydot 3, this file is missing. This is probably caused by #380 where the license file has been renamed and moved into a `LICENSES` folder. It looks like this folder is not considered when building the wheel.

ferdnyc:
Add a `[tool.setuptools]` section to `pyproject.toml` containing a `license-files` array that globs for all of our license files, to ensure they end up in the built wheel.

The files end up at the root of the `.dist-info` directory, not in a `LICENSES` subdirectory, but I don't know how to change that and it seems "fine". At least they're there.

Fixes #390

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ build-backend = "setuptools.build_meta"
 name = 'pydot'
 description = "Python interface to Graphviz's Dot"
 readme = "README.md"
-license = {file = "LICENSE"}
+license = {text = "MIT"}
 requires-python = ">= 3.8"
 dependencies = [
   'pyparsing>=3.0.9'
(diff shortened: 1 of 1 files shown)


ferdnyc:
@lkk7 This is effectively the same as the `setup.cfg` change you tested, but uses `pyproject.toml` to keep as much of the configs in one place as possible. (I'd love to get rid of `setup.cfg` entirely, if only `tox` and `tox-gh` had better `pyproject.toml` support.)

lkk7 (approved):
OK, looks good!

PR #394, opened 2024-07-19
Use temporary directories for temporary file output

ferdnyc:
This PR modernizes the handling of temporary files, switching to the `tempfile.TemporaryDirectory()` context manager in both `pydot.core.Dot.write()`, and the `test_dot_args()` unit test (which previously created a `test.svg` file in whatever directory the tests were run from).

Using the context manager means that temporary files will be cleaned up in _all_ cases, even when exceptions are raised or things exit unexpectedly. I have a _lot_ of `/tmp/tmpXXXXXX` files containing graphviz code on my system, to prove that this wasn't happening reliably before.

There is some danger using `tempfile.TemporaryDirectory()` on Windows systems, which can encounter permissions errors during cleanup. An `ignore_cleanup_errors` argument was introduced in Python 3.10 to handle those, but isn't available in older releases. We'll have to see how the Windows CI runs turn out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,7 +51,6 @@ coverage.xml
 *.cover
 .hypothesis/
 .pytest_cache/
-test.svg
 
 # Translations
 *.mo
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -1776,44 +1776,34 @@ class Dot(Graph):
             args = []
 
         # temp file
-        tmp_fd, tmp_name = tempfile.mkstemp()
-        os.close(tmp_fd)
-        self.write(tmp_name, encoding=encoding)
-        tmp_dir = os.path.dirname(tmp_name)
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            fp = tempfile.NamedTemporaryFile(dir=tmp_dir, delete=False)
+            fp.close()
+            self.write(fp.name, encoding=encoding)
 
-        # For each of the image files...
-        for img in self.shape_files:
-            # Get its data
-            f = open(img, "rb")
-            f_data = f.read()
-            f.close()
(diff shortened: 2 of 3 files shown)


ferdnyc:
> There is some danger using `tempfile.TemporaryDirectory()` on Windows systems, which can encounter permissions errors during cleanup. An `ignore_cleanup_errors` argument was introduced in Python 3.10 to handle those, but isn't available in older releases. We'll have to see how the Windows CI runs turn out.

Huh! No troubles at all. :+1:

lkk7:
So, the CI passes, but is there a possibility of this giving errors for some Windows users?

ferdnyc:
@lkk7 

> So, the CI passes, but is there a possibility of this change causing errors for some Windows users?

I can't definitively claim there isn't, but the fact that the CI doesn't trigger any issues is a very good sign, since it should exercise all of the possibilities.

The main concern would be that some files are left open somewhere, either by us or the user, and therefore can't be deleted. I tried to be very careful about that, in our own code. And we don't ever pass filehandles or open files out to users, so they don't have to be. (Either they provide a path _not_ in the temp directory that we write to, or we read the tempfile ourselves and it gets closed when we return the in-memory data.) So our tempdirs are pretty well isolated, and there shouldn't be much danger of user code being able to interfere with the cleanup.

ferdnyc:
(The other worry would be that we copy files (the shape files) into the tempdir that have permissions which prevent deletion. But we don't ever use `shutil.copy2()` or `shutil.copyfile()`, we read from each input file and write the data to a newly-created filehandle inside the temp space, so it will be created with default permissions instead of potentially carrying along the source file's.)

lkk7 (approved):

PR #407, opened 2024-08-17
Warn about thread safety where appropriate

### LINKED ISSUE #401: Data Race in parse_dot_data.
in function `parse_dot_data`,  a global variable `top_graphs` is used. This global variable will be overwritten repeatedly in multithreading situations.  It is a data race and may produce unexpected results.

https://github.com/pydot/pydot/blob/1270c6d8e7e77e8dbdad460d22d6448658893717/src/pydot/dot_parser.py#L499-L516

The following is an example used with `ThreadPoolExecutor` that may cause data race if the user does not know the implementation of pydot(the call chain is `graph_from_dot_file` -> `graph_from_dot_data` -> `parse_dot_data`):
```python
import pydot
from tqdm import tqdm
from concurrent.futures import ThreadPoolExecutor, as_completed

with ThreadPoolExecutor(max_workers=max_workers) as executor:
    futures = [executor.submit(pydot.graph_from_dot_file,  dot_file) for dot_file in file_list]
    for future in tqdm(as_completed(futures), total=len(futures))
        future.result()
```

lkk7:
Closes #401

### DIFF
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -497,6 +497,9 @@ def graph_definition():
 def parse_dot_data(s):
     """Parse DOT description in (unicode) string `s`.
 
+    This function is NOT thread-safe due to the internal use of `pyparsing`.
+    Use a lock if needed.
+
     @return: Graphs that result from parsing.
     @rtype: `list` of `pydot.Dot`
     """
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -405,6 +405,9 @@ def quote_attr_if_necessary(s):
 def graph_from_dot_data(s):
     """Load graphs from DOT description in string `s`.
 
+    This function is NOT thread-safe due to the internal use of `pyparsing`.
+    Use a lock if needed.
+
     @param s: string in [DOT language](
         https://en.wikipedia.org/wiki/DOT_(graph_description_language))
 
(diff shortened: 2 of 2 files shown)


src/pydot/dot_parser.py
```diff
@@ -497,6 +497,9 @@ def graph_definition():
 def parse_dot_data(s):
     """Parse DOT description in (unicode) string `s`.
 
+    This function is NOT thread-safe due to the internal use of `pyparsing`.
+    Use a lock if needed.
+
     @return: Graphs that result from parsing.
     @rtype: `list` of `pydot.Dot`
     """
```
ferdnyc, line 505: Makes sense to me!

At some point we might think about reformatting all of the docstrings into Sphinx format — or, better, [sphinx one-line format](https://github.com/NilsJPWerner/autoDocstring/blob/c9da64126fd9e667decd9d85b4e5b53c60372ea7/docs/one-line-sphinx.md) (the difference being that Sphinx format uses separate `:param foo:` and `:type foo:` lines, while one-line format is `:param \<type> foo:`), then we could make these into "Warning" admonitions. e.g.:

```python
def parse_dot_data(s):
    """Parse DOT description in (unicode) string `s`.

    :return list[pydot.Dot]: Graphs that result from parsing.

    .. warning:: This function is NOT thread-safe due to
       the internal use of :mod:`pyparsing`.
       Use a :class:`threading.Lock` if needed.
    """
```

But it makes more sense to do that all at once, in a separate PR. IMHO.

ferdnyc (approved):

ferdnyc:
Oh, fabulous, those CI failures are the Windows tempfile issues I was hoping #394 weren't going to cause.

I'll figure something out, maybe we can wrap `tempfile.TemporaryDirectory` and implement our own `ignore_cleanup_errors` for Python < 3.10.

PR #409, opened 2024-08-19
Vendor in Python 3.12 tempfile.py, for < 3.10

ferdnyc:
Python 3.8 and 3.9 are missing the `ignore_cleanup_errors` option to `tempfile.TemporaryDirectory()`, which is needed to ensure that spurious Windows tempfile cleanup errors don't lead to crashes.

Include a copy of the Python 3.12 `tempfile.py` (slightly modified to be compatible with Python 3.8 `types` and `shutil.rmtree`), and use that instead of the standard library module to ensure that we can use `ignore_cleanup_errors` in all Python versions.

When our minimum supported version is Python 3.10, we can change `from pydot._vendor import tempfile` back to `import tempfile` and delete `src/pydot/_vendor/tempfile.py`.

### DIFF
--- a/LICENSES/Python-2.0.txt
+++ b/LICENSES/Python-2.0.txt
@@ -1,0 +1,72 @@
+PYTHON SOFTWARE FOUNDATION LICENSE VERSION 2
+
+     1. This LICENSE AGREEMENT is between the Python Software Foundation ("PSF"), and the Individual or Organization ("Licensee") accessing and otherwise using this software ("Python") in source or binary form and its associated documentation.
+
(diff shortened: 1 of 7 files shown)


ferdnyc:
Hmm, now I have to figure out how to ignore paths in Ruff.

ferdnyc:
No API changes, but still we should _proooobably_ push out a 3.0.2 release once this is merged, to ensure that nobody using Pydot on Windows hits these same failures.

lkk7 (approved):

ferdnyc:
It's insane that those failures _just_ started happening out of the blue, and even MORE insane is the file they were failing on. Consistently, every time (modulo the Python version and the randomized name of the directory under `Temp`):

```text
    File "C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\shutil.py",
 line 618, in _rmtree_unsafe
      os.unlink(fullname)
  PermissionError: [WinError 5] Access is denied:
'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpgwhk_3lj\\
%SystemDrive%\\ProgramData\\Microsoft\\Windows\\Caches\\
cversions.2.db'
```

wat


Sounds like it might be [the frickin' _search indexer_](https://www.windows10forums.com/threads/what-is-c-programdata-microsoft-windows-caches-folder-for.17514/) creating them, which is just... ... ... peak Microsoft, let's just leave it at that.

PR #412, opened 2024-08-22
Experimentally create separate ruff-check/fix envs

ferdnyc:
Since tox's `{posargs}` can't be used to transform a `ruff format --diff` call into a `ruff format` call, split the `ruff` env into a `ruff-check` env for the CI system, and a (non-default) `ruff-fix` env that can be run manually using `tox -e ruff-fix` to have changes auto-applied.

I _think_ this will work out with the CI setup (and `ruff-fix` will just be ignored because it isn't on any list of default environments, so it should be manual-only), but it'll be good to see how the first run goes.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,7 +17,7 @@ env_list =
     py310
     py39
     py38
-    ruff
+    ruff-check
 
 [testenv]
 extras = tests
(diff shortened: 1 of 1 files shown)


ferdnyc:
And I'm just now realizing... would it be better to have `ruff` run _before_ the CI job, when it's included in the environment set?

Right now we have `3.12 = py312, ruff` (or now `ruff-check`) in the tox-gh mappings, which means it'll run through all the tests first, _then_ run the `ruff` checks.

But if `ruff` is going to fail a PR on code-formatting grounds, why not get that out of the way **first**, _then_ run the tests? Always better to know sooner, and it's not like the tests won't still be running in like 17 other CI jobs at the same time.

ferdnyc:
That feels better to me, somehow. It's a difference of a whopping 17 seconds (the runtime of the test suite, these days, on `ubuntu-latest`), but it still feels right-er.

lkk7 (approved):

lkk7:
Nice. 

On a side note, could you look at the very end of this comment? https://github.com/common-workflow-language/cwltool/pull/2019#issuecomment-2299734889

ferdnyc:
@lkk7

I have the GitHub app notification for that conversation still haunting my phone, because I keep meaning to get to it and don't want to dismiss it or I'll **definitely** forget...

PR #414, opened 2024-08-28
Exclude _vendor from coverage and display badge in README

ferdnyc:
Get the uninteresting `_vendor/` code out of the coverage data, since it's not our job to test Python standard library classes.

Have the README display the coverage badge from the `python-coverage-comment-action-data` branch. (It's autogenerated from the action runs on `main` commits — so, currently 77%, but it'll be 86% after this merges — and stored at `/badge.svg` on the branch.)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -15,4 +15,9 @@ source =
   test
 source_pkgs =
   pydot
+omit =
+  **/_vendor/*
 
+[report]
+omit =
+  **/_vendor/*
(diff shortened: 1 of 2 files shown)


ferdnyc:
> 
> _This PR does not seem to contain any modification to coverable code._

Maybe not, but...

>  

Hey-hey! :tada: Effects of booting out the `_vendor` subdir.

lkk7 (approved):
Great, let's see the effect on the README.md.

ferdnyc:
Clicking on that badge, BTW, takes you to the HTML coverage report produced by `coverage.py`, which is also hosted in the action's dedicated branch. From there you can browse coverage-annotated views of all the source files.

PR #420, opened 2024-09-20
Enhance parser and name-lookup with better quoting

### LINKED ISSUE #399: nx_pydot.graphviz_layout(G, prog='dot') does not accept nodes with spaces
When switching to the new version of pydot==3.0.1, I noticed a change of behaviour of this part of my code : <br>
`pos = nx.nx_pydot.graphviz_layout(G, prog='dot')` where G is a simple nx.MultiDiGraph. <br>Instead of returning me the pos, it returned me this error : `line 263, in graphviz_layout
    return pydot_layout(G=G, prog=prog, root=root) > line 339, in pydot_layout
    node = node[0]
IndexError: list index out of range`.<br> I noticed that it only happened when one of my MultiDiGraph node was named with a space inside. I'm not sure it is a normal behaviour but I'm posting this here in case it is not.

### LINKED ISSUE #413: The pydot `dot_parser`, and round-tripping through `dot`
So, I've finally (belatedly) had a chance to wrap my brain around #396 better. (cc: @lkk7)

The reason for all that conversion, in the networkx code, is that they were round-tripping their `pydot.Graph` object through the `dot` command and reading it back in again, which is actually a fairly common technique in the Graphviz world.

### Graphviz layout techniques
The reason for doing it is, having `dot` _output_ a `dot`-format file (`dot -Tdot`) tells it to take the input graph definition(s) **and write them back out with all of the implicit elements made explicit, and the layout applied** in the form of a bounding-box (`bb=`) attribute applied as a default `graph` attribute, and `height=`, `width=`, and `pos=` attributes applied to all of the elements.

For example, this graph:

```dot
graph G {
  a [shape=octagon, color=red];
  b [shape=triangle, color=blue];
  a -- b [dir=both];
  b -- { c; d; };
}
```
...passed through `dot -Tdot`, will come out like this:

```dot
graph G {
	graph [bb="0,0,126,194.5"];
	node [label="\N"];
	{
		c	[height=0.5,
			pos="27,18",
			width=0.75];
		d	[height=0.5,
			pos="99,18",
			width=0.75];
	}
	a	[color=red,
		height=0.5,
		pos="63,176.5",
		shape=octagon,
		width=0.75];
	b	[color=blue,
		height=0.70139,
		pos="63,97.25",
		shape=triangle,
		width=0.97428];
	a -- b	[dir=both,
		pos="s,63,158.36 e,63,122.97 63,147.03 63,142.84 63,138.47 63,134.15"];
	b -- c	[pos="57.375,84.18 51.284,71.11 41.529,50.178 34.723,35.572"];
	b -- d	[pos="68.625,84.18 74.716,71.11 84.471,50.178 91.277,35.572"];
}
```

That's valuable information for anyone who wants to then **modify** that layout in any way, particularly to do things Graphviz won't do on its own.

#### WireViz example
For example, it's an integral part of the technique we hit upon for the WireViz tool, which wanted to [post-process edges representing wires in its graphs, to route them _through_ table nodes](https://forum.graphviz.org/t/straitening-one-line-throu-a-table/2196/24). WireViz uses Graphviz to output a layout where edges are connected from the `:w` port of a node to the `:e` port of the same node. Graphviz will naturally lay that edge out so it loops arooooound the outside of the node. But, by taking the start and end coordinates in that edge's `pos=` data, a post-processing script can adjust the `pos=` values to route the edge _through_ the node, instead.

For example, this code:
```dot
graph G {
  "W" [shape=box, label=<
<table width="100%" border="0" cellspacing="0" cellborder="0">
  <tr><td>&nbsp;</td><td>&nbsp;</td><td>&nbsp;</td></tr>
  <tr><td colspan="3" cellpadding="0" port="w1">
    <table width="100%" border="0" cellspacing="0" cellpadding="0">
      <tr><td colspan="3" cellpadding="0" height="2" bgcolor="#000000"></td></tr>
      <tr><td colspan="3" cellpadding="0" height="2" bgcolor="#ff0000"></td></tr>
      <tr><td colspan="3" cellpadding="0" height="2" bgcolor="#00ff00"></td></tr>
      <tr><td colspan="3" cellpadding="0" height="2" bgcolor="#ff0000"></td></tr>
    </table>
  </td></tr>
</table>
>];
W:w1:w -- W:w1:e [style=bold, color="#000000:#ff0000:#00ff00:#000000"];
}
```

Rendered by `neato` will normally create a loopy edge:

wv

But by running it through `neato -Tdot`, then adjusting the `pos=` values for the edge and processing it _again_ with `neato -n2` (the `-n2` activates neato's highest level of "no-layout" mode, where it will skip its node-positioning and edge-routing steps and just respect the supplied `pos=` values), the graph can be turned into this:

wv2

### Pydot support
So, like I said, it's a common and useful technique. So common, in fact, that we might consider adding special support for it in `pydot` — perhaps a method `pydot.Dot.layout()` that sends the graph data through `<command> -Tdot`, then automatically runs it back through the parser and populates a new `pydot.Graph`, with all of the elements now annotated with layout attributes.

But whether or not we do **that**, the crux of the issue is that pydot's `Graph.to_string()` and `dot_parser.parse_dot_data()` don't combine to be idempotent for identifiers — passing a graph definition through one followed by the other gets you graph elements that don't have the same node names as the original data. 

In pydot 3.0.0 we changed the logic underpinning the graph element constructors so they preserved the formatting of identifiers suppplied as arguments — `pydot.Node("a 1")` now creates a node named `a 1` instead of a node named `"a 1"`, for example. But when the output of `pydot.Graph.to_string()` is run through `pydot.dot_parser.parse_dot_data()`, that same guarantee is no longer present — `pydot.Node("a 1")` becomes `"a 1"` in the graph source, which `dot_parser` then parses back in as `"a 1"`, and creates the corresponding node by calling `pydot.Node("\"a 1\"")`.

While the results of parsing a graph round-tripped through `dot -Tdot` will never be **exactly** the same as the input — observe, for example, the breaking out of the implicit subgraph in the graph above, along with the separation of `a -- { c d };` into `a -- c` and `a -- d`. But there are things we could do to make the effects _less_ disruptive to  the input data than our current code will be.

### Potential quoting changes in Pydot
Things we could do to make round-tripping less likely to confuse calling code are:

#### On output: Stop replacing `\n` with `\\n` and `\r` with `\\r`

There's really no reason to do that. Graphviz fully supports embedded newlines in quoted graph identifiers. This code:

```dot
graph G {
  "a\nnode" [color=red];
}
```

Is exactly the same as _this_ code:
```dot
graph G {
  "a
node" [color=red];
}
```

in terms of how it's treated by the Graphviz parser.

Our parser also supports capturing raw newlines inside quoted IDs:

```python
>>> import pydot
>>> source = '''graph G {
...   "a
... node" [color=red];
... }'''
>>> g = pydot.dot_parser.parse_dot_data(source)[0]
>>> g.get_nodes()[0].get_name()
'"a\nnode"'
```

So, there's no reason to turn `\n` into `\\n`, or `\r` into `\\r`, inside `.to_string()`.

#### On input: Have the parser suppress quotes around values

What bit the networkx folks in #396 was the transformation of `pydot.Node("a\nb")` into `pydot.Node('"a\\nb"')`, which is the fault of both `.to_string()` _and_ `dot_parser`. While `.to_string()` caused `\n` => `\\n`, the **parser** caused the ID value to become wrapped in double quotes.

To more consistently match the behavior of our API calls (post-3.0.0), `dot_parser` should be _suppressing_ the outer quotes around quoted IDs when it reads them in, so that `"a b"` becomes `pydot.Node("a b")`, not `pydot.Node('"a b"')`.

That'd also be more in line with Graphviz's own parser, which will for example strip off unnecessary quotes around IDs:

```console
$ echo 'graph G { "a" [color=red]; }' | dot -Tdot
graph G {
	graph [bb="0,0,54,36"];
	node [label="\N"];
	a	[color=red,
		height=0.5,
		pos="27,18",
		width=0.75];
}
```

Our own parser will read that in as `'"a"'`, leaving the unnecessary quotes as part of the ID value:

```python
>>> import pydot
>>> g = pydot.dot_parser.parse_dot_data(
... 'graph G { "a" [color=red]; }')[0]
>>> g.get_nodes()[0].get_name()
'"a"'
```

If the parser were to suppress quotes when parsing, and `Graph.to_string()` continue to only re-add them on output (and only when necessary) the way it already does since 3.0.0, we gain the ability to parse the output of `pydot.Graph.to_string()` without the identifiers being changed in the process. 

Which means we get closer to this code:

```python
>>> import pydot
>>> g = pydot.Graph("G", graph_type="graph")
>>> g.add_node(pydot.Node("foo bar"))
```
..and this code...:
```python
>>> import pydot
>>> g = pydot.dot_parser.parse_dot_data(
... 'graph G { "foo bar"; }')[0]
```
being functionally equivalent.

### LINKED ISSUE #436: dot_parser.py: Why does ID not have a `float_number` the `float_number` production rule?
From the [DOT language reference](https://graphviz.org/doc/info/lang.html):

> An ID is one of the following:
> 
>     Any string of alphabetic ([a-zA-Z\200-\377]) characters, underscores ('_') or digits([0-9]), not beginning with a digit;
>     a numeral [-]?(.[0-9]⁺ | [0-9]⁺(.[0-9]*)? );
>     any double-quoted string ("...") possibly containing escaped quotes (\")¹;
>     an HTML string (<...>).
> 
> An ID is just a string; the lack of quote characters in the first two forms is just for simplicity. There is no semantic difference between abc_2 and "abc_2", or between 2.34 and "2.34". Obviously, to use a keyword as an ID, it must be quoted.

I see that it is omitted from `ID` [dot_parser.py](https://github.com/pydot/pydot/blob/a892962a2db1a71f5e0aa83cfa734720ce2bb077/src/pydot/dot_parser.py#L403) and there is instead a righthand ID that satisfies the complete definition. Is there a particular reason for this?

```python
ID = (identifier | html_text | double_quoted_string).setName("ID")
```

ferdnyc:
This is Pydot 4.0 work for sure, since its changes are pretty sweeping

## Stripping quotes during parsing

In #413 I mentioned that it would enhance round-tripping of our graphs through `dot` and back again if the parser _unquoted_ input strings where possible. This PR starts the work of doing that.

Because not all strings can be unquoted, instead of enabling `unquote_results` in `pyparsing.QuotedString()`, a new parse action `possibly_unquote` is attached to the `double_quoted_string` parser element that handles making that decision. It makes sure that any quoted strings stay quoted that need to: double-quoted strings containing what would otherwise parse as an HTML-like string, or as an ID with port values, etc. If it will confuse the pydot code when sent in unquoted, it's left quoted. Otherwise, it's unquoted, which matches `dot`'s own behavior in parsing.

In addition, the _lookup_ functions `get_node()`, `get_subgraph()`, and `get_edge()` are enhanced to also try the _quoted_ versions of their ID arguments, if passed unquoted strings.<sup>1</sup> So, we're better protected both coming and going.

The `make_quoted` utility function will no longer translate newlines and carriage returns into `\n` and `\r` — it's unnecessary, as `dot` quoted strings can contain the raw control characters without issue, and our parser can consume them as well. So, leaving them be will further enhance our ability to round-trip graphs through graphviz commands with better fidelity.

Fixes #413

### Notes
1. For `get_edge()` in particular, that became extremely complex, because each endpoint has to be separately checked for a possible quoted form, and tested with or without a quoted form of the _other_ endpoint. 
   So, for example, `get_edge('abc', 'def')` will try:
   * In a directed graph,
       1. `('abc', 'def')`
       2. `('"abc"', '"def"')`
   * In an undirected graph,
       1. `('abc', 'def')`
       2. `('"abc"', '"def"')`
       3. `('def', 'abc')`
       4. `('"def"', '"abc"')`
   
   But when one side is already quoted, the quoted form of the other side still has to be tested as well. So, `get_edge('<<i>html</i>>', 'def')` will try:
   * In a directed graph,
       1. `('<<i>html</i>>', 'def')`
       2. `('<<i>html</i>>', '"def"')`
   * In an undirected graph,
       1. `('<<i>html</i>>', 'def')`
       2. `('<<i>html</i>>', '"def"')`
       3. `('def', '<<i>html</i>>')`
       4. `('"def"', '<<i>html</i>>')`
   
   That code is ugly, but I dropped in a long-ass explanatory comment which will hopefully help.

## Improved numeric parsing in `dot_parser`

1. Negative-value floats will be recognized as such — previously, we didn't allow a leading `-` sign.
2. Floats are included in the grammar for `ID` values, and are legal unquoted in all places where graphviz allows them.
3. The parser now correctly requires that unquoted identifiers _start with_ a non-digit, meaning an alpha character or `_`. This was already true in the quoting logic, but now the parser does the right thing too.
4. The idea of a "right-hand ID" (that is, an ID _after_ an equals sign, in an assignment) is eliminated — IDs are IDs are IDs. There was never supposed to be any difference in the grammar.

The upshot of all those changes is that, in `main`:

```python
>>> import pydot.dot_parser as dp; gp = dp.graph_definition()
>>> # This was previously an error
>>> g = gp.parse_string(
...     'graph G { -5 [color=red, length=-4.5, width=5]; }')[0]
>>> # And this was previously...
>>> g = gp.parse_string(
...     'graph G { "-5" [color=red, length="-4.5", width=5]; }'
... )[0]
>>> print(g.to_string())
graph G {
"-5" [color=red, length="-4.5", width=5];
}
```

On this branch:
```python
>>> import pydot.dot_parser as dp; gp = dp.graph_definition()
>>> g = gp.parse_string(
...     'graph G { -5 [color=red, length=-4.5, width=5]; }'
... )[0]
>>> print(g.to_string())
graph G {
-5 [color=red, length=-4.5, width=5];
}
>>> g = gp.parse_string(
...     'graph G { "-5" [color=red, length="-4.5", width=5]; }'
... )[0]
>>> print(g.to_string())
graph G {
-5 [color=red, length=-4.5, width=5];
}
```

Much more in line with graphviz.

Due to these changes, this PR now...

Fixes: #436 

## Move constants to `pydot.constants`

A new file `constants.py` is added to `pydot`, containing all of the `CONSTANT_LIST_VARIABLES` previously defined in `pydot.core`. (The actual reason for this move is, I also need the `DOT_KEYWORDS` list in `dot_parser` now.)

Because all of the constants in `pydot.constants` are currently imported into `pydot.core`, the `__init__.py` file's `from pydot.core import *` import will still pick them up, which means that things like `pydot.NODE_ATTRIBUTES` and `pydot.OUTPUT_FORMATS` will still be defined even after this change.

However, I'm tempted to change that (e.g. by having `pydot.core` use `from pydot import constants`, and refer to the constants as `constants.CONSTANT_NAME`), in order to get the constants lists out of the root `pydot` namespace. They were never really intended to be referenced from the `pydot.` level.


ferdnyc:
Aargh, Well, that failure's for a completely different reason: I forgot you can't do this in Python 3.8:

```python
with (
    open(some_file) as in1,
    open(some_file) as in2
):
```

(It can be done, but it can't be wrapped in parens and made multi-line until Python 3.10.)

ferdnyc:
Oh! Turns out, the `b53.dot` failure was due to the missing `.` quoting from #419, since I started that branch before this was merged. As soon as I rebased this on master, the tests went back to all-green.

This still shouldn't be merged until we're sure we're ready to start development on Pydot 4.0.0, but it's no longer a WIP as it's now passing all tests, same as `main`. (All adjusted tests, since the expected results of some of the API tests has to be modified for the new `get_foo()` logic.)

ferdnyc:
Oh, right, but I still can't do this in Python < 3.10.

```python
with (
    open(some_file) as in1,
    open(some_file) as in2
):
```

ferdnyc:
This is my least favorite ruff change of all time:

```diff
ruff-check: commands[0]> ruff format --diff .
--- src/pydot/core.py
+++ src/pydot/core.py
@@ -264,7 +264,7 @@
 
 def make_quoted(s: str) -> str:
     """Transform a string into a quoted string, escaping internal quotes."""
-    out = s.replace('"', r'\"')
+    out = s.replace('"', r"\"")
     return f'"{out}"'
```

ferdnyc:
I just added a number of commits that I was working on as a separate PR, but realized they only make sense in the context of this one, since it doesn't help to improve the parsing of floats (which this does) without also improving their output, and vice versa.

With the new commits,

1. Negative-value floats will be recognized as such — previously, we didn't allow a leading `-` sign.
2. Floats are included in the grammar for `ID` values, and are legal unquoted in all places where graphviz allows them.
3. The parser now correctly requires that unquoted identifiers _start with_ a non-digit, meaning an alpha character or `_`. This was already true in the quoting logic, but now the parser does the right thing too.
4. The idea of a "right-hand ID" (that is, an ID _after_ an equals sign, in an assignment) is eliminated — IDs are IDs are IDs. There was never supposed to be any difference in the grammar.

The upshot of all those changes is that, in `main`:

```python
>>> import pydot.dot_parser as dp; gp = dp.graph_definition()
>>> # This was previously an error
>>> g = gp.parse_string(
...     'graph G { -5 [color=red, length=-4.5, width=5]; }')[0]
>>> # And this was previously...
>>> g = gp.parse_string(
...     'graph G { "-5" [color=red, length="-4.5", width=5]; }'
... )[0]
>>> print(g.to_string())
graph G {
"-5" [color=red, length="-4.5", width=5];
}
```

On this branch:
```python
>>> import pydot.dot_parser as dp; gp = dp.graph_definition()
>>> g = gp.parse_string(
...     'graph G { -5 [color=red, length=-4.5, width=5]; }'
... )[0]
>>> print(g.to_string())
graph G {
-5 [color=red, length=-4.5, width=5];
}
>>> g = gp.parse_string(
...     'graph G { "-5" [color=red, length="-4.5", width=5]; }'
... )[0]
>>> print(g.to_string())
graph G {
-5 [color=red, length=-4.5, width=5];
}
```

Much more in line with graphviz.

ferdnyc:
I also eliminated all of the "punctuation" variables in the parser (things like `comma = Literal(",")` and so on); it's clearer to just use the actual character wherever you mean to. So this:

```python
port = (
    Group(Group(colon + ID) + Group(colon + ID))
    | Group(Group(colon + ID))
).setName("port")
```

became this:
```python
port = (
    Group(Group(":" + ID) + Group(":" + ID)) | Group(Group(":" + ID))
).setName("port")
```

Where the punctuation was being suppressed (`lbrace.suppress()`), the pyparsing Suppress class is used instead: `Suppress("{")`.

ferdnyc:
Note to self: Make double-sure that the parsing logic doesn't _unquote_ the node names in this graph:

```dot
graph G {
  "graph" [color=red];
  "edge" -- "graph";
}
```

That would be bad. (In fact, it's worth including a test for that particular case.)

ferdnyc:
Glad I thought to check that. It gets it right for edges, but screws up nodes. So, to draft status it goes.

src/pydot/dot_parser.py
```diff
@@ -383,7 +385,7 @@ def push_node_stmt(s: str, loc: int, toks: ParseResults) -> "pydot.core.Node":
     return n
 
 
-class HTML(Token):
+ class HTML(Token):
```
ferdnyc, line 388: No idea what happened here.
```suggestion
class HTML(Token):
```

ferdnyc:
I think this might be done. Even if there's more I _could_ do, it's big enough already that I'd rather save anything further for a separate PR.

Oh, I need to add tests for the whole make-sure-the-parser-doesn't-unquote-keyword-names thing I spotted, and fixed, above. Which I'd rather put in `test_parser.py`, so I think I'll wait until #452 is merged (which creates that file).

ferdnyc:
I think this branch is now good, but I'm not sure how reviewable it is anymore. I may have to re-build it on top of the current `main` branch `HEAD`, to make it comprehensible.

lkk7:
Yes, you can create an issue, or I'll create it, whatever. Thanks!

lkk7:
If you ever continue this PR, just a small ping for this one: https://github.com/common-workflow-language/cwltool/pull/2019#issuecomment-2299184385

It's possible that the current port quoting is too "eager". Or it isn't, but it's just an unfortunate improvement that broke things.

PR #423, opened 2024-10-05
Move/rename pydot.core.frozendict to pydot.classes.FrozenDict, clean up, bugfix, and fully test

ferdnyc:
This PR pulls the `frozendict` class out of `core.py`, creating a new `classes.py` file to hold it, and gives it the PEP8-compliant class name `FrozenDict`. Along the way, it's significantly cleaned up, streamlined, tweaked, and fully tested (every line covered except the `__repr__` method).

* All of the disabled methods are fully defined, with the proper argument signatures, instead of all being aliased to `_blocked_method`.
* The `__new__` method is streamlined and made more robust, with one initialization bug fixed
* Key sorting is removed from the `__hash__()` method, and `__eq__` / `__ne__` methods are defined so that FrozenDicts with differently-ordered members will test as not equal.
* A full set of basic unit tests cover every line of the `pydot.classes` module except for `FrozenDict.__repr__` (including testing that all of the disabled methods properly raise `AttributeError` on attempts to modify the `FrozenDict`).
* `pydot.core.frozendict` is still defined, as an alias for `pydot.classes.FrozenDict` which creates a `DeprecationWarning` on use.

### DIFF
--- a/src/pydot/classes.py
+++ b/src/pydot/classes.py
@@ -57,7 +57,7 @@ class FrozenDict(dict):
     def __new__(cls, *args, **kw):
         new = dict.__new__(cls)
         args_ = [cls._freeze_arg(arg) for arg in args]
-        dict.__init__(new, *args_, **kw)
+        dict.__init__(new, *args_, **cls._freeze_arg(kw))
         return new
 
     def __init__(self, *args, **kw):
--- a/src/pydot/classes.py
+++ b/src/pydot/classes.py
@@ -63,6 +63,14 @@ class FrozenDict(dict):
     def __init__(self, *args, **kw):
         pass
 
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            return hash(self) == hash(other)
+        return NotImplemented
+
+    def __ne__(self, other):
+        return not self == other
+
     def __hash__(self):
         try:
             return self._cached_hash
--- a/src/pydot/classes.py
+++ b/src/pydot/classes.py
@@ -1,0 +1,75 @@
+# SPDX-FileCopyrightText: 2024 pydot contributors
+#
+# SPDX-License-Identifier: MIT
+
+"""Frozen dictionaries."""
+
+import copy
+
+
+class FrozenDict(dict):
+    """Frozen dictionary, values are immutable after creation.
+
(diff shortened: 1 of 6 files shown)


ferdnyc:
I'm bummed that, even with these changes, because of the "sneaky" way `FrozenDict` is created from a `dict` instance, I wasn't able to change any of these unfortunate properties of the class:

```python
>>> import pydot
>>> fd = pydot.FrozenDict((('a', 1), ('b', 2)))
>>> isinstance(fd, dict)
True
>>> import collections.abc
>>> isinstance(fd, collections.abc.MutableMapping)
True
```

It would be preferable for both of those to return `False`, but `isinstance(fd, collections.abc.Mapping)` still return `True`. But unfortunately, being a subclass of `dict` means that `FrozenDict` is treated as a `dict` by the type system — having its mutation methods overridden with ones that raise `AttributeError` doesn't change that. (And the methods can't be deleted from `FrozenDict`, because then it will fall back to the methods from its parent `dict` class which **do** mutate the object.)

ferdnyc:
The subclassing of `dict` at the heart of `FrozenDict` also means that `FrozenDict() == dict()` will fall back to using `dict.__eq__()` for the equality test, meaning that `FrozenDict({'a': 1, 'b': 2}) == dict({'b': 2, 'a': 1})` will return `True` even though `FrozenDict({'a': 1, 'b': 2}) == FrozenDict({'b': 2, 'a': 1})` will (now) return `False`.)

lkk7:
Yeah, I hope no one depends on these details in their application logic 😄

src/pydot/classes.py
```diff
@@ -0,0 +1,83 @@
...
+
+class FrozenDict(dict):
+    """Frozen dictionary, values are immutable after creation.
+
+    Extended version of ASPN's Python Cookbook Recipe:
+    https://code.activestate.com/recipes/414283/
+
+    This version freezes dictionaries used as values within dictionaries."""
+
+    _block_msg = "A frozendict cannot be modified."
```
ferdnyc, line 18: I think I hate myself a little bit, for not fixing the "FrozenDict" capitalization here. :roll_eyes:

PR #435, opened 2024-11-05
Implement .resolve() method for attributes

ferdnyc:
This WIP PR implements a method <code>.resolve(</var>name</var>)</code> for all graph element classes, which can be thought of as "`.get()` with defaults".

Where <code>.get(<var>name</var>)</code> will only look up the value for the <var>name</var> attribute if it's set on the object for which it's called, <code>.resolve(<var>name</var>)</code> will look for that attribute, and if it's not found, walk up the graph hierarchy scanning for a relevant default value. If one is found, that value is returned instead.

Internally, a `.find_default(name, element_type, seq_end)` method on Graph handles walking the hierarchy. It will consider the attributes of any `node`, `edge`, or `graph` nodes that correspond to the type of the current element, and also have lower sequence numbers. (Nodes are checked in reverse sequence order, first one with a matching attribute returns the value.) Any direct graph attributes on the parent graph(s) are also checked, when looking up subgraph attributes.

### Remaining work
- [ ] An open question: **Should** this be a separate method, or should it be an optional feature of `.get()`? e.g. something like <code>.get(<var>name</var>, find_default=True)</code>
- [ ] Tests. There are none, yet, and obviously this is something that needs to be tested very thoroughly to make sure the default-finding logic matches graphviz's own.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -1130,7 +1130,9 @@ class Graph(Common):
         self.obj_dict["current_child_sequence"] += 1
         return seq
 
-    def find_default(self, name: str, element_type: str, seq_end: int = -1) -> Any:
+    def find_default(
+        self, name: str, element_type: str, seq_end: int = -1
+    ) -> Any:
         """Find the default value for attribute 'name' on 'element_type'.
 
         If 'seq_end' is set, use it to limit the range of sequence numbers
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -573,6 +573,16 @@ class Common:
         """
         return self.obj_dict["attributes"].get(name, None)
 
+    def resolve(self, name):
+        """Resolve an attribute value, with defaults applied.
+
+        If the attribute 'name' is directly set on the graph element,
+        return its value. Otherwise, walk up the hierarchy and look for
+        any relevant default values for that attribute; return the first
+        one found.
+        """
+        return self.get(name)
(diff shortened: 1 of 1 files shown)


lkk7:
> An open question: Should this be a separate method, or should it be an optional feature of .get()? e.g. something like .get(name, find_default=True)

Imagining myself as a random user, I see `get(..., find_default=True)` as easier to understand than `resolve(...)`. But maybe it's just the naming – "find default" is clearer than "resolve". On the other hand, `find_default(...)` as a method by itself seems wrong, because it seems like it will ONLY find the default value.

ferdnyc:
@lkk7 

> Imagining myself as a random user, I see `get(..., find_default=True)` as easier to understand than `resolve(...)`.

\*nod\* Or maybe `use_defaults` is a better argument name, even. (It's certainly even more implicitly boolean, name-wise.) But I do like the idea of improving `.get()` rather than creating a whole new method.

(I'd also have to patch the same functionality into all of the generated getters, but that's easy enough to do, and it would mean that `.get_foo()` remains functionally equivalent to `.get("foo")`, since I was never planning to generate a whole new pile of `.resolve_foo()` methods.)

In theory I could even enhance `.get_attributes(use_defaults=True)` to act as a "smart" version of `get_*_defaults` that would incorporate _all_ of the reachable defaults applicable to the element in question.

PR #437, opened 2024-11-20
Simplify get_top_graph_type with default

ferdnyc:
To make use of get_top_graph_type simpler for callers, add an optional argument "default", with a default value of "graph". This value will be returned if the element has no identifiable top graph type (e.g. an Edge not yet added to a Graph).

"graph" was chosen as the default (even though the _actual_ default type for graphs is "digraph") because it generally triggers more expansive processing in code that relies on it. The tools to look up edges, for example, will use the endpoints in both orderings, when in an undirected "graph" context.

This also makes the typing easier, as `get_top_graph_type` now returns `str` in all cases.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -12,7 +12,7 @@ import re
 import subprocess
 import sys
 import warnings
-from typing import Any, List, Optional, Sequence, Set, Tuple, Type, Union
+from typing import Any, List, Optional, Sequence, Set, Tuple, Type, Union, cast
 
 import pydot
 import pydot.dot_parser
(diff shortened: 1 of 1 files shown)


ferdnyc:
> "graph" was chosen as the default (even though the _actual_ default type for graphs is "digraph") because it generally triggers more expansive processing in code that relies on it.

I forgot to mention that it's _also_ the default we were using anyway. In `main`:

```python
>>> import pydot
>>> e = pydot.Edge('a', 'b', color='red')
>>> e.to_string()
'a -- b [color=red];'
>>> g = pydot.Graph("G")
>>> g.add_edge(e)
>>> e.to_string()
'a -> b [color=red];'
```

PR #439, opened 2024-11-27
Fix "all numeric" quoting behavior

### LINKED ISSUE #438: Pydot considers dotted strings like "1.4.0" to be numeric
@pktpls wrote:

> Have had the same problem with version numbers in node properties, which end up as e.g. `ipkVersion=1.4.0`, while they should actually be wrapped in quotes

Ooh, that's a different problem. The regular expression used to detect numeric values accepts periods (because, decimal points), but really it should only accept a SINGLE period at most. We'll have to change that from 

```python
re_all_numeric = re.compile(r"^[0-9\.]+$")
```

to something like...

```python
re_all_numeric = re.compile(r"^([0-9]+\.?[0-9]*|[0-9]*\.[0-9]+)$")
```

_Originally posted by @ferdnyc in https://github.com/pydot/pydot/issues/178#issuecomment-2504157116_

ferdnyc:
The `re_all_numeric` check used to allow any number of decimal points in the value, which isn't correct for numerics; at most one can be allowed. Rename to `re_numeric`, and use a more complex RE that allows a maximum of one decimal point, but fails on strings like "1.4.0".
    
Fixes #438

A new unit test confirms the correct behavior for numeric strings with varying numbers of decimal points.


ferdnyc:
@lkk7 

It might make sense to backport this onto the v3.0.2 branch, and release a v3.0.3 with this fix, since I don't know that 4.0 is ready to go out the door yet. LMK if you'd like me to put together another PR for that.

lkk7:
Thanks, I'll try to release v3.0.2 and v3.0.3 tomorrow or this weekend.

PR #440, opened 2024-11-28
Backport #439 to v3.0

ferdnyc:
The fixes from #439, cherry-picked and targeting the v3 branch.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -293,7 +293,7 @@ def make_quoted(s):
 
 dot_keywords = ["graph", "subgraph", "digraph", "node", "edge", "strict"]
 
-re_all_numeric = re.compile(r"^[0-9\.]+$")
+re_numeric = re.compile(r"^([0-9]+\.?[0-9]*|[0-9]*\.[0-9]+)$")
 re_dbl_quoted = re.compile(r'^".*"$', re.S)
 re_html = re.compile(r"^<.*>$", re.S)
 
@@ -322,7 +322,7 @@ def any_needs_quotes(s):
     if has_high_chars and not re_dbl_quoted.match(s) and not re_html.match(s):
         return True
 
-    for test_re in [re_all_numeric, re_dbl_quoted, re_html]:
+    for test_re in [re_numeric, re_dbl_quoted, re_html]:
         if test_re.match(s):
             return False
 
--- a/test/test_pydot.py
+++ b/test/test_pydot.py
@@ -275,6 +275,12 @@ class TestGraphAPI(PydotTestCase):
         self.graph_directed.add_node(pydot.Node(1))
         self.assertEqual(self.graph_directed.get_nodes()[0].get_name(), "1")
 
+    def test_numeric_quoting(self):
+        num = pydot.Node("12", label=1.4, width=".75")
+        non = pydot.Node("1.2.3", label="1.4.0")
+        self.assertEqual(num.to_string(), "12 [label=1.4, width=.75];")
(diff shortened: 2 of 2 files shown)


lkk7:
@ferdnyc advice is needed!

3.0.3 is released.
I don't have much experience with release management - I wonder if I should rebase the current `main` onto the `v3` branch so that the former fully contains the latter.
Because right now, the `main` branch doesn't contain the `3.0.3` tag , while `v3` does, which doesn't sit right with me.
Or maybe I'm wrong and we should treat them as separate from now on. But it's odd that the `v3.0.3` tag doesn't show up [here](https://github.com/pydot/pydot/tags) on GitHub, probably because it only shows tags from the `main` branch.

ferdnyc:
@lkk7

> @ferdnyc advice is needed!
> 
> 3.0.3 is released. I don't have much experience with release management - I wonder if I should rebase the current `main` onto the `v3` branch so that the former fully contains the latter. 

I definitely wouldn't rebase `main`, you'd end up in conflict hell for every commit since the `v3` branch point. Merging `v3` back _into_ `main` might be more of an option. But I'm not convinced that's necessary.

> Because right now, the `main` branch doesn't contain the `3.0.3` tag , while `v3` does, which doesn't sit right with me. 

Does it? Because, I can't find any evidence of a `v3.0.3` tag **anywhere**, on any branch/commit, either in the web UI or in my local checkout. Even after having fully fetched the project repo and all tags. `git tag` shows me up to `v3.0.2`, but not `v3.0.3`.

Did you create the tag locally on the command line? Git won't push tags by default — you either need to `git push <remote> v3.0.3` or `git push [<remote>] --tags`, to have it push a local tag up to the project repo.

> Or maybe I'm wrong and we should treat them as separate from now on. But it's odd that the `v3.0.3` tag doesn't show up [here](https://github.com/pydot/pydot/tags) on GitHub, probably because it only shows tags from the `main` branch.

It's possible that'll be true even when the tag is pushed — [this section of the docs](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/about-commits#about-commit-branches-and-tag-labels) makes some vague noises that _might_ imply that would be the case. But it's not super clear:

> If your commit is not on the default branch (main), the label will show the branches which contain the commit. If the commit is part of an unmerged pull request, you can click the link to go to the pull request.
> 
> Once the commit is on the default branch, any tags that contain the commit will be shown and the default branch will be the only branch listed. For more information on tags, see "[Git Basics - Tagging](https://git-scm.com/book/en/v2/Git-Basics-Tagging)" in the Git documentation.

I know that, in many git workflows, releasing is done in branches that are _never_ returned to the default branch, because the release commits they contain (setting the release version number, for example) aren't part of the development workflow. Or, sometimes, projects maintain multiple branches indefinitely for different release series, so after branching off the development tree each follows its own, divergent path. That can certainly work with _git_. I just don't know if it's possibly incompatible with GitHub.

I'd try to make sure that the `v3.0.3` tag exists in the repo first, even if it's not on the `main` branch. I **think** that should be OK. If not, then I guess we cross that bridge.

lkk7:
Of course you're right. All this time I thought I had committed the tag to the repository. But now that it's actually there, GitHub shows the tag and everything is fine!

ferdnyc:
@lkk7 

I'm sure there's a good reason git doesn't push tags by default... well, let me rephrase that. I'm sure there's a reason git doesn't push tags by default, and I _assume_ someone thought it was a **good** reason when they made that decision, but it sure is annoying.

PR #443, opened 2024-12-06
Improve initialization, getters

ferdnyc:
Some cleanup of the main classes' setup and access methods:

### Init `obj_dict` in Common, and use `super()`
Give the Common class an `__init__()` that sets up the `obj_dict`, and call `super().__init__` from all of the subclasses. Makes the code in the subclass `__init__`s simpler.

### Simplify graph getters
    
* Calls like `get_node_defaults()` and `get_strict()` took an extra second argument that was never used for anything — eliminate it.
    
* The `get_*_defaults()` methods also wasted time checking whether the return value of `get_node()` is a list, but it's always a list.

### Standardize some naming
Don't call a variable `dict` (use `_dict` instead), and use `*args` and `**kwargs` in function signatures, not other weird names.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -1001,13 +1001,9 @@ class Graph(Common):
     def set_graph_defaults(self, **attrs: Any) -> None:
         self.add_node(Node("graph", **attrs))
 
-    def get_graph_defaults(self, **attrs: Any) -> Any:
+    def get_graph_defaults(self) -> Any:
         graph_nodes = self.get_node("graph")
-
-        if isinstance(graph_nodes, (list, tuple)):
-            return [node.get_attributes() for node in graph_nodes]
-
-        return graph_nodes.get_attributes()
+        return [node.get_attributes() for node in graph_nodes]
 
     def set_node_defaults(self, **attrs: Any) -> None:
         """Define default node attributes.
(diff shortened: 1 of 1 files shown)


ferdnyc:
Note that the `get_*_defaults` methods are completely untested, and arguably not that useful. Here's how `get_node_defaults()` works, for example:

```python
>>> import pydot
>>> g = pydot.Graph("G", graph_type="graph")
>>> g.add_node(pydot.Node("node", color="red"))
>>> g.add_node(pydot.Node("node", shape="box", color="blue"))
>>> g.get_node_defaults()
[{'color': 'red'}, {'shape': 'box', 'color': 'blue'}]
```

It just returns a list of all the `node`-node attribute dictionaries. For some reason!

I assume the additional arguments were meant to supply default-defaults, so that you could do something like this:

```python
>>> g.get_node_defaults(penwidth=2)
```

...and get back a set of attributes with either the default penwidth, or `2` if none was specified. But that was never implemented, nor was any code combining the attributes of multiple default nodes, when some are overridden.

All in all, the concept of "defaults" isn't an absolute, it's dependent on the position an element occupies within the graph definition tree — the default color for node 'A', here, is different from the default color for node 'B':

```dot
graph G {
  node [color=red];
  A [shape=box];
  node [color=blue];
  B [shape=circle];
}
```

So, given that the way they work makes no sense, and they're completely untested (0 coverage), it's likely nobody uses the `{get,set}_*_defaults` Graph methods at all. It may be better to just eliminate them.

One possible replacement would be my `.resolve()` enhancement (see #435), which is a method on individual objects, and _does_ take into account the position of the given element. `node_a.resolve("color")` would return `"red"`, while `node_b.resolve("color")` would return `"blue"`.

lkk7 (approved):

PR #451, opened 2025-01-07
Treats HTML inside < and > as a single token

### LINKED ISSUE #235: Parser breaks off lines at hash symbol following HTML tag
This issue is split off from issue #203. We found out [there](https://github.com/pydot/pydot/issues/203#issuecomment-534135106) that the parser breaks off lines inside DOT HTML-like labels containing `<BR/>#`. It breaks them off starting from the hash/pound sign/number sign (`#`).

Here are two minimal reproducible examples:

**Example 1** shows that everything starting from `#` to the end of the line is dropped:

```python
import pydot
G = pydot.graph_from_dot_data("""
graph G {
    b1 [label=<
         We are the knights who say <BR/># Ni
    >];
}
""")
print(G[0])
```

Output:
```dot
graph G {
b1 [label=<
         We are the knights who say <BR/>
    >];
}
```
Notice that `# Ni` is missing.

**Example 2** shows that this can become a parsing error when parts with a syntactic meaning are lost:

```python
import pydot
G = pydot.graph_from_dot_data("""
graph G { b4 [label=<We are the knights who say <BR/># Ni>]; }
""")
```

Output:
```
graph G { b4 [label=<We are the knights who say <BR/># Ni>]; }
             ^
Expected "}", found '['  (at char 14), (line:2, col:14)
```
These three lines of output are actually the "explanation" of the `ParseException` that is raised by PyParsing. Although the input line is printed whole as part of the explanation, the parser currently does not see further than the `#` character and therefore loses sight of the closing brackets `>]; }` at the end. That is why it says it expected `}`, but could not find it.

Obviously, the `#` is considered the start of a comment, but there is more to it than that, because that does not happen in all cases. Other tests I did seem to indicate that the word-by-word parsing, using whitespace as delimiters, plays a role as well. I suspect that in the end our construction of the DOT Language in PyParsing terms in `dot_parser` will need to be tweaked to fix this bug.

I will try to post more details later as I find time. If someone else wants to dive into this, please let me know so that we won't be doing the same work twice.

Versions used for examples: Python 3.7.3, pydot 1.4.1+PR227, pyparsing 2.4.7.

davidweichiang:
The `HTML` token scans until all `<` are matched with `>`, even if there are `#` characters inside.

Closes #235.

### DIFF
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -30,7 +30,6 @@ from pyparsing import (
     Token,
     Word,
     cStyleComment,
-    lineno,
     nums,
     pyparsing_unicode,
     restOfLine,
@@ -404,7 +403,13 @@ class HTML(Token):
             loc += 1
             if num_open == 0:
                 return loc, instring[open_loc:loc]
-        raise ParseException(instring, loc, f"expected a > to match <, in the HTML string starting at {lineno(open_loc, instring)}", self)
+        raise ParseException(
+            instring,
+            loc,
+            "expected a > to match <, in the HTML string"
+            + "starting at {lineno(open_loc, instring)}",
+            self,
+        )
 
 
 graphparser = None
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -17,7 +17,6 @@ import typing as T
 
 from pyparsing import (
     CaselessLiteral,
-    CharsNotIn,
     Combine,
     Forward,
     Group,
@@ -383,25 +382,29 @@ def push_node_stmt(s: str, loc: int, toks: ParseResults) -> "pydot.core.Node":
     n = pydot.Node(str(node_name), **attrs)
     return n
 
-class HTML(Token):
-    def __init__(self):
-        super().__init__()
 
-    def parseImpl(self, instring, loc, do_actions=True):
+class HTML(Token):
+    def __init__(self) -> None:
+        super().__init__()  # type: ignore
+
+    def parseImpl(
+        self, instring: str, loc: int, do_actions: bool = True
+    ) -> T.Tuple[int, str]:
         start = loc
-        if not (loc < len(instring) and instring[loc] == '<'):
+        if not (loc < len(instring) and instring[loc] == "<"):
             raise ParseException(instring, loc, "expected <", self)
         open = 1
         loc += 1
         while loc < len(instring):
-            if instring[loc] == '<':
+            if instring[loc] == "<":
                 open += 1
-            elif instring[loc] == '>':
+            elif instring[loc] == ">":
                 open -= 1
             loc += 1
             if open == 0:
                 return loc, instring[start:loc]
         raise ParseException(instring, loc, "expected >", self)
+
 
 graphparser = None
 
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -30,6 +30,7 @@ from pyparsing import (
     Token,
     Word,
     cStyleComment,
+    lineno,
     nums,
     pyparsing_unicode,
     restOfLine,
@@ -390,20 +391,20 @@ class HTML(Token):
     def parseImpl(
         self, instring: str, loc: int, do_actions: bool = True
     ) -> T.Tuple[int, str]:
-        start = loc
+        open_loc = loc
         if not (loc < len(instring) and instring[loc] == "<"):
             raise ParseException(instring, loc, "expected <", self)
-        open = 1
+        num_open = 1
         loc += 1
         while loc < len(instring):
             if instring[loc] == "<":
-                open += 1
+                num_open += 1
             elif instring[loc] == ">":
-                open -= 1
+                num_open -= 1
             loc += 1
-            if open == 0:
-                return loc, instring[start:loc]
-        raise ParseException(instring, loc, "expected >", self)
+            if num_open == 0:
+                return loc, instring[open_loc:loc]
+        raise ParseException(instring, loc, f"expected a > to match < at {lineno(open_loc, instring)}", self)
 
 
 graphparser = None
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -28,6 +28,7 @@ from pyparsing import (
     ParserElement,
     ParseResults,
     QuotedString,
+    Token,
     Word,
     cStyleComment,
     nums,
@@ -382,6 +383,25 @@ def push_node_stmt(s: str, loc: int, toks: ParseResults) -> "pydot.core.Node":
     n = pydot.Node(str(node_name), **attrs)
     return n
 
+class HTML(Token):
+    def __init__(self):
+        super().__init__()
+
+    def parseImpl(self, instring, loc, do_actions=True):
+        start = loc
+        if not (loc < len(instring) and instring[loc] == '<'):
+            raise ParseException(instring, loc, "expected <", self)
+        open = 1
+        loc += 1
+        while loc < len(instring):
+            if instring[loc] == '<':
+                open += 1
+            elif instring[loc] == '>':
+                open -= 1
+            loc += 1
+            if open == 0:
+                return loc, instring[start:loc]
+        raise ParseException(instring, loc, "expected >", self)
 
 graphparser = None
 
@@ -418,12 +438,7 @@ def graph_definition() -> ParserElement:
             '"', multiline=True, unquoteResults=False, escChar="\\"
         )
 
-        html_text = Forward()
-        inner_html = OneOrMore(CharsNotIn("<>") | html_text)
-        html_text << "<" + inner_html + ">"
-        html_text.setParseAction(lambda arr: "".join(arr))
-
-        ID = (identifier | html_text | double_quoted_string).setName("ID")
+        ID = (identifier | HTML() | double_quoted_string).setName("ID")
 
         float_number = Combine(
             Optional(minus) + OneOrMore(Word(nums + "."))
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -404,7 +404,7 @@ class HTML(Token):
             loc += 1
             if num_open == 0:
                 return loc, instring[open_loc:loc]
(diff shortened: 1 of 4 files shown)


ferdnyc:
Thanks @davidweichiang!

> even if there are `#` characters inside.

Do comment chars break the parsing, even despite the `OneOrMore(CharsNotIn("<>"))`? ...I suppose that makes sense, in its way.

(Perhaps there's a "more Pyparsing-ic" way to make that "greedier", so it gloms up **ALL** characters that aren't the matching close-`>`, including (even) comment chars — but that's ultimately what your `HTML` class is doing anyway, so meh.)

I very much hate the idea of _actually parsing_ the contents of HTML-like strings **at all**, and we've studiously avoided it at every turn so far.

This seems to strike the right balance of achieving the goal, while still avoiding interpretation of the actual contents as much as possible, so that's a big plus.

(Interestingly, the [Graphviz Visual Editor](https://magjac.com/graphviz-visual-editor/) makes the same parsing mistake and _syntax-highlights_ `# comments` inside an HTML-like label as if they're comments — even though the text is clearly interpreted as part of the label:)

image

ferdnyc:
@davidweichiang Looks like your `class HTML(Token)` needs type annotations, which is making mypy unhappy, and Ruff has a bunch of issues, as it always does.

Easiest approach would probably be to make the type changes first (if you want to test locally, `tox -e mypy-check` will run mypy on the code), then when mypy is happy, you can run `tox -e ruff-fix`, let it modify whatever it needs to, and commit the results.

davidweichiang:
> Do comment chars break the parsing, even despite the OneOrMore(CharsNotIn("<>"))? ...I suppose that makes sense, in its way.

Actually I don't know why the `ignore` gets inside of the `OneOrMore(CharsNotIn("<>"))`. I'll dig into that further.

And I'll fix the type annotations.

src/pydot/dot_parser.py
```diff
@@ -382,6 +383,25 @@ def push_node_stmt(s: str, loc: int, toks: ParseResults) -> "pydot.core.Node":
...
+        if not (loc < len(instring) and instring[loc] == '<'):
+            raise ParseException(instring, loc, "expected <", self)
+        open = 1
+        loc += 1
+        while loc < len(instring):
+            if instring[loc] == '<':
+                open += 1
+            elif instring[loc] == '>':
+                open -= 1
+            loc += 1
```
ferdnyc: We're so _close_ to being able to do something smarter here (but can't), with the tracking of `<` characters.

In actual Graphviz HTML-like _code_, it's impossible to have a tag nested inside another tag (meaning, not as a child, but _between_ its `<` and its `>`. So, in **theory**, `open` should never be more than `2`.

**EXCEPT!** inside _comments_. That's what ruins it. There, basically anything goes, as long as your angle braces balance out in the end.

```gv
graph G {
  good [label=<
    Totally legal: <!-- <<<<>>>> -->
    Also A-OK:     <!-- <><><><><><><> -->
    No probs here: <!-- <<<><>><<<>>>><<>> -->
  >];
  bad [label=<
    Boom, parse error: <!-- <<> -->
  >];
}
```

...If it weren't for that, we could abort parsing the moment `open` becomes `3`. But, alas.

I do wonder, though, if it might be worth keeping track of things a bit more, for error-reporting purposes.

What if instead of merely counting `<`s, we pushed its location onto a stack each time we saw one, then popped it whenever we encounter a `>`? That way, if we reach the end of the string and the stack isn't empty, instead of just reporting "Expected `>`" at `loc`, we could report something like "Missing close `>` for `<` at `[open_loc]`" at `loc`.

(We'd probably want to convert `[open_loc]` from a character index to a line and column number. I assume the Pyparsing API provides a means of doing that, tho as I was writing this last night I got sidetracked digging into _how_ and never finished it. — This comment, I mean. Tho I never found the answer to my question, either, so really both.)

If that's too complex, we could also just raise `ParseException("Unclosed <", open_loc)` and let Pyparsing handle it.

test/my_tests/hash_inside_html_1.dot
```diff
@@ -0,0 +1,6 @@
+// https://github.com/pydot/pydot/issues/235
+graph G {
+    b1 [label=<
+         We are the knights who say <BR/># Ni
+    >];
+}
```
ferdnyc, line 6: I think if we're going to test this, we should also test the use of HTML-like strings as an identifier, in addition to a label. We can do it in the same file, but I'd add a second test, something like...

```suggestion

    <<b>b2 #rules!</b>> [color=red];
}
```

ferdnyc (approved):
Modulo the typing annotations and Ruff being made happy, looks good! I've left some _suggestions_, but nothing so critical that it would prevent merging this. Thanks, @davidweichiang !

ferdnyc:
@davidweichiang 

> Or a separate lexing pass (just for comments, quoted strings, and HTML) ought to work too.

I think that's definitely overkill. The `NoIgnore` thing is an _option_, but I don't imagine we'd ever use it anywhere **else** so generalizing that way may not be worth it. It's up to you.

I think your `HTML()` class is a perfectly good solution. Arguably better, as it's more specific about its purpose. "NoIgnore" tells future-us much less about what it's doing and (more importantly) _why_ it's doing it.

src/pydot/dot_parser.py
davidweichiang: In real HTML I thought `<!-- <<> -->` should be okay, but GraphViz indeed gives an error, so the `HTML` class is doing the right thing.

davidweichiang: I've changed the error message as suggested. However, pyparsing catches the exception and the actual error message is something else (`Expected graph_stmt, found '['  (at char 15), (line:2, col:6)`).

test/my_tests/hash_inside_html_1.dot
davidweichiang: I'm really surprised to learn that that is valid DOT but I've added the test!

src/pydot/dot_parser.py
```diff
@@ -383,6 +383,29 @@ def push_node_stmt(s: str, loc: int, toks: ParseResults) -> "pydot.core.Node":
...
+        loc += 1
+        while loc < len(instring):
+            if instring[loc] == "<":
+                open += 1
+            elif instring[loc] == ">":
+                open -= 1
+            loc += 1
+            if open == 0:
+                return loc, instring[start:loc]
+        raise ParseException(instring, loc, "expected >", self)
```
ferdnyc: The uncovered line here is fine; we probably _should_ have more failure-tests, but at the moment we don't have that many.

And I wouldn't even want to go too hard into populating them as unittest `assertRaises` calls, until we finally get all of the testing rewritten in pytest.

src/pydot/dot_parser.py
```diff
@@ -383,6 +384,29 @@ def push_node_stmt(s: str, loc: int, toks: ParseResults) -> "pydot.core.Node":
...
+        loc += 1
+        while loc < len(instring):
+            if instring[loc] == "<":
+                num_open += 1
+            elif instring[loc] == ">":
+                num_open -= 1
+            loc += 1
+            if num_open == 0:
+                return loc, instring[open_loc:loc]
+        raise ParseException(instring, loc, f"expected a > to match < at {lineno(open_loc, instring)}", self)
```
ferdnyc, line 407: Mmm, almost, but the mismatched `<` isn't _necessarily_ the first one we encountered.

I was thinking something more like this:

```suggestion
        if not (loc < len(instring) and instring[loc] == "<"):
            raise ParseException(instring, loc, "expected <", self)
        open_locs = [loc]
        pos = loc + 1
        while pos < len(instring):
            if instring[pos] == "<":
                open_locs.append(pos)
            elif instring[pos] == ">":
                open_locs.pop()
            pos += 1
            if len(open_locs) == 0:
                break
        if len(open_locs) > 0:
            unmatched = open_locs[-1]
            raise ParseException(
                instring,
                pos,
                f"expected a > to match < at {lineno(unmatched, instring)}",
                self)
        return cursor, instring[loc:pos]
```

...Is probably about how Ruff would want to format it.

(Part of me thinks the `if(): break` should come _before_ incrementing `pos` that one last time, but I'm not sure so I'll trust you on that one.)

ferdnyc: Or, come to think of it, if you prefer the `while` condition can just be:

```python
while len(open_locs) > 0 and pos < len(instring):
```

ferdnyc: ^ I just updated my suggestion to add that, and corrected another issue that would've come up if `open_locs` _emptied_ and we hit another `>`.

davidweichiang: Since this error message seems to never see the light of day, I thought it would just be simpler to rephrase it so it's accurate.

src/pydot/dot_parser.py
```diff
@@ -383,6 +384,29 @@ def push_node_stmt(s: str, loc: int, toks: ParseResults) -> "pydot.core.Node":
     return n
 
 
+class HTML(Token):
+    def __init__(self) -> None:
+        super().__init__()  # type: ignore
```
ferdnyc, line 389: BTW, that linked PR (pyparsing/pyparsing#596) is meant to fix things so we can eventually drop this `type: ignore` comment.

src/pydot/dot_parser.py
ferdnyc: @davidweichiang 

Yeah, I see what you mean. Hrm. And if it's a node name, you get:

```text
Expected graph_stmt, found '<' (at char 10), (line:1, col:11)
```

That's disappointing. I suppose our parser would have to be a lot more stateful, to surface these inner-rule errors. Pyparsing's `MatchFirst` (aka `|`) is very forgiving; whenever the input doesn't match a rule it just assumes it's something else and moves on.

Oh, well. Makes sense to me, then.

But that error message isn't going to fly, Ruff will complain about the too-long line.

Since nobody's going to see it, I wouldn't worry about making it too detailed. Dropping back to `Expected >` would even be fine, IMHO. (Sorry for the runaround. my fault!)

lkk7:
Thanks for the change! If there are no more updates/corrections planned for this PR, then I'll sanity-check and merge it soon.

lkk7:
I checked it earlier, just forgot to merge. Thanks!

PR #452, opened 2025-01-28
Update dot_parser.HTML and add tests

ferdnyc:
I'm an idiot.

And, apologies to @davidweichiang for my being an idiot.

If I'd _not_ been an idiot, I'd have realized that **of course** we can test the `HTML` parsing class, including its exception — just, not within the context of the complete `graphparser`. We test it just like we'd unit test any other class: _In isolation._

If I'd realized that, and written (or asked for) some unit tests on the class, we'd have caught that:

1. The `pyparsing.lineno` import went missing. (Ruff removed it.)
2. ...because the exception message for the `HTML` class was missing the `f` on its (non-)f-string, which means the `{lineno()}` call wasn't being interpreted as Python code.
3. And it was also missing a space between two words (where it was split into two strings).

So, this PR fixes those things, and adds a new test file `test/test_parser.py`, which currently contains only some pytest tests for `pydot.dot_parser.HTML` — both successful and failed parses. (The failed parse is where we can actually see its exception in use.)

It also moves the class up to the top of the `dot_parser.py` file with the other classes, gives it a docstring, and shortens the exception message slightly (as it proved a bit unwieldy, in practice).

### DIFF
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -30,6 +30,7 @@ from pyparsing import (
     Token,
     Word,
     cStyleComment,
+    lineno,
     nums,
     pyparsing_unicode,
     restOfLine,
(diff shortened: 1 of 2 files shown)


lkk7:
... Yeah, that shows how much my "sanity check" is worth, especially when I wasn't fully focused on it. Thanks for the fix.

PR #453, opened 2025-01-29
Port API tests to pytest

ferdnyc:
This PR makes the following changes to the `test/` directory:

- Remove unittest class `TestGraphApi` from `test_pydot.py`
- Create new `test_api.py`, containing tests from `TestGraphApi` suite ported to pytest
- Add `graph_directed` fixture to `conftest.py`
- Create new `test_logging.py` to host `test_logging_init` test (only) -- it seems to makes the tests run cleaner

Along the way, it fixes two bugs, one in the tests, one in the library:

1. One of the tests was attempting to look up nodes with `.get_node('"name"')`, which in pydot 3.0+ is no longer necessary unless the node was created with a quoted name.
2. One of the pickling tests contained a call of the form
   ```python
   pydot.Graph.add_edge(pydot.Edge(("D", "E")))
   ````
   which we never noticed didn't actually work as expected. The resulting graph threw a traceback when its `.to_string()` was called, because the edge added was from `("D", "E") -- ""`, which couldn't be stringified properly by `.to_string()` (because one of its endpoints was a tuple, not a string).

    The Edge constructor will now accept a list or tuple as the first arg, and create an edge between the (presumably two... tho it doesn't actually check) items in the sequence.


lkk7:
Thanks!

PR #454, opened 2025-01-29
tests: Also output dot sources, on failure

ferdnyc:
A change I pulled out of #420 (to try and slim it down and make it a bit more focused):

If `TEST_ERROR_DIR` is set when running the tests, and a failure is encountered in one of the file-driven tests, write out both the original file and the `.to_string()` for the parsed version, for comparison.

### DIFF
--- a/test/test_pydot.py
+++ b/test/test_pydot.py
@@ -12,6 +12,7 @@
 import functools
 import os
 import sys
+import typing as T
 import unittest
 from hashlib import sha256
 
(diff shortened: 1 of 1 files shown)


ferdnyc:
Yeah, that's the downside to adding more failure-handling code to the tests — the coverage takes a hit as long as tests _aren't_ failing.

I'm tempted to add an intentional failure (marked XFAIL of course) just to exercise the code that handles exporting test-failure artifacts and keep it from bitrotting.

Maybe after/during the pytest-porting of those last file-based test classes.

lkk7 (approved):

PR #461, opened 2025-03-13
Tests: Eliminate a bunch of redundant test files

ferdnyc:
Going through the `test/graphs/` directory, I discovered that it contained:

1. A PostScript file (`dfd.ps`) that was never referenced anywhere in the repo; deleted.
2. Six pairs of files that were byte-for-byte identical; deleted one of each pair.
3. Three file-groups that were too similar to all be useful:
   - `xx.dot` and `b102.dot` were identical except for the font size value of each element; deleted `xx.dot`.
   - `lsunix[123].dot` were wildly (and unhelpfully) similar; deleted `lsunix[12].dot` and renamed `lsunix3.dot` to `lsunix.dot`.
   - `process.dot` and `p[234].dot` were also iterations of the same graph, kept only the most complex (`p4.dot`)


lkk7:
I've never ventured into those old files, thank you!

PR #464, opened 2025-04-04
dot_parser: Define parser elements in a class

ferdnyc:
This PR is centered around one major, should-be-api/user-transparent change to the `dot_parser` code which will change everything for us:

The final `ParserElement` that we use for parsing, **as well as all of the component parsing fragments used to build it**, are now defined in the top level of a class definition named `GraphParser`.

The class is not meant to be instantiated (though it can be), and it consists only of class-member variables, all of which are the `ParserElement` components of our parser definition. The final element, `GraphParser.parser`, is what used to be stored in the global `graphparser` symbol. (Which is preserved for compatibility, even though our own code won't use it anymore.)

Having access to the parser components is a necessary step towards being able to finally debug, test, and modify the parser. With the new class, I can do things like this:

```python3
>>> from pydot.dot_parser import GraphParser

>>> # If I want to test our parsing of attribute lists...
>>> GraphParser.attr_list.parse_string(
... '''[color="red", style="dashed, filled", width=12.5]''')
ParseResults([P_AttrList({'color': '"red"', 'style': '"dashed, filled"', 'width': '12.5'})], {})

>>> # Or edge statements...
>>> elist = GraphParser.edge_stmt.parse_string(
... '''a:5:sw -- b:4:ne [color=blue]''')
>>> print(elist[0].to_string())
a:5:sw -- b:4:ne [color=blue];

>>> # Or if I want to start trying to figure out why our
>>> # parsing of complex edges is totally broken...
>>> g = GraphParser.parser.parse_string(
... '''graph G { a -- { b; c; } -- d; }''')
>>> print(g[0].to_string(indent=2))
graph G {
  a -- subgraph {
    b;
    c;
  };
}

>>> # Well, that's clearly not right. What is the edge
>>> # parser doing?
>>> for e in GraphParser.edge_stmt.parse_string(
... '''a -- b -- c [color=blue]'''):
...     print(e.to_string())
... 
a -- b [color=blue];
b -- c [color=blue];

>>> # So far so good, but...
>>> GraphParser.edge_stmt.parse_string(
... '''a:5:sw -- { n; m } -- b:4:ne [color=blue]''')
ParseResults([<pydot.core.Edge object at 0x7fcc1089e5d0>], {})
>>> elist = GraphParser.edge_stmt.parse_string(
... '''a:5:sw -- { n; m } -- b:4:ne [color=blue]''')
>>> elist[0].to_string()
'a:5:sw -- subgraph {\nn;\nm;\n} [color=blue];'
```

So now I at least have a starting point to look into the code. It's also possible to disable parse actions (just call `.set_parse_action()` with no arguments on any of the `ParserElement` objects that have one, to clear it out) in order to get a look at the values being sent _into_ the parse action.

And we can write unit tests for all of the parser pieces, now, instead of only testing it as a black box.

## Other changes made
- `dot_parser` now only imports, and only refers to, `pydot.core.*` classes; the compatibility exports at the `pydot` level are never used. That makes the dependency graph more manageable.
- By the same token, the top-level import of `pydot.dot_parser` is gone from `core.py`. It's now only imported on demand, when `graph_from_dot_data()` needs it.
- All type hints in `dot_parser` are now real references to imported classes, it no longer uses string annotations.
- Unused `s` and `loc` arguments are removed from parse action functions that don't use them (something PyParsing has supported for a long while, now).
- <s>The no-op `__init__` is removed from the `HTML` class, since its unused type-ignore comment was annoying MyPy and it served no purpose anyway.</s> OK, correction, `HTML.__init__()` can't be removed, just its `# type: ignore` comment.

## Notes
This will cause huge conflicts with my other open PR, but it's worth the disruption. I'll do the work of resolving those once they occur.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,5 +61,5 @@ python =
     3.12 = py312
     3.11 = py311
     3.10 = py310
-    3.9 = py39
-    3.8 = mypy-check, py38
+    3.9 = mypy-check, py39
+    3.8 = py38
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -15,7 +15,6 @@ import warnings
 from typing import Any, List, Optional, Sequence, Set, Tuple, Type, Union, cast
 
 import pydot
-import pydot.dot_parser
 from pydot._vendor import tempfile
 from pydot.classes import AttributeDict, FrozenDict
 
@@ -401,6 +400,8 @@ def graph_from_dot_data(s: str) -> Optional[List["Dot"]]:
     @return: Graphs that result from parsing.
     @rtype: `list` of `pydot.Dot`
     """
+    import pydot.dot_parser
+
     return pydot.dot_parser.parse_dot_data(s)
 
 
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -544,3 +544,7 @@ def parse_dot_data(s: str) -> T.Optional[T.List[pydot.core.Dot]]:
         print(" " * (err.column - 1) + "^")
         print(err)
         return None
+
+
+# Backwards compatibility
+graphparser: ParserElement = GraphParser.parser
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -82,7 +82,7 @@ class HTML(Token):
     """Parsing for HTML-like strings."""
 
     def __init__(self) -> None:
-        super().__init__()  # type: ignore
+        super().__init__()
 
     def parseImpl(
         self, instring: str, loc: int, do_actions: bool = True
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -36,7 +36,7 @@ from pyparsing import (
     restOfLine,
 )
 
-import pydot
+import pydot.core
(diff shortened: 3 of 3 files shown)


ferdnyc:
So, I just added another commit that moves the lower-end `mypy-check` test run from the Python 3.8 environment to the Python 3.9 environment.

### Rationale

The code as it currently stands _passes_ MyPy checks on Python 3.9 - 3.13, it only upsets MyPy on Python 3.8 because PyParsing is untyped on that version, and always will be as it's no longer supported by newer PyParsing.

And since Python 3.8 is _also_ EOL, while we may still **support** it, I don't think we should write our code **for** that version.

If the code _works_ on Python 3.8 (and it does), and if it passes type checks on _supported_ Python versions (which it does), then I think that's the highest level of Python 3.8 support we can be expected to provide.

lkk7 (approved):
Great work! I read it twice and didn't see anything to complain about.
It passes the tests.

ferdnyc:
Now I just have to take advantage of this newfound power and _actually_ figure out why `push_edge_stmt` is so very, VERY broken.

And also possibly hack in support for the undocumented syntax I found last week (#465).

PR #467, opened 2025-04-11
Fix complex edge parsing

ferdnyc:
This PR makes minimal changes to the parser to fix the parsing of what I'm calling "complex" edges, meaning ones that contain subgraph endpoints. The parser is now able to correctly interpret a graph with edges like these:

```dot
graph G {
  a -- {b; c;} -- d -- {e; f; };
  {x [color=blue]; y [color=red]; z; } -- {m n o p} [penwidth=5];
}
```

Specifically, after parsing that becomes:

```dot
graph G {
  a -- {
    b;
    c;
  };
  {
    b;
    c;
  } -- d;
  d -- {
    e;
    f;
  };
  {
    x [color=blue];
    y [color=red];
    z;
  } -- {
    m;
    n;
    o;
    p;
  } [penwidth=5];
}
```
(Note to self: Fix the `inline` mode in `Subgraph.to_string()`...)

And renders to SVG as:

g

A new test file `test/graphs/complex_edges.dot` is added to exercise the feature and ensure that our output is the same as `dot`'s.

### Other changes

* A preceding setup commit eliminates all of the `.setName()` calls on `ParserElement`s in the `GraphParser` definition, by instead calling the `autoname_elements()` function from PyParsing when finished. This is a feature that's been available for some time, that sets the name of every local `ParserElement` to the name of the variable it's stored in. Which is exactly what we were doing anyway.

* The eventual definition of `Forward()` `ParserElement`s (which is now only `stmt_list`, I managed to eliminate the unnecessary forward-declaration of `edge_point`) now uses `<<=` instead of `<<`, another PyParsing adjustment since `<<` isn't meant to assign values to things. Abusing it to do so made static analysis messier.

### DIFF
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -18,6 +18,7 @@ import typing as T
 from pyparsing import (
     CaselessLiteral,
     Combine,
+    DelimitedList,
     Forward,
     Group,
     Literal,
@@ -351,50 +352,26 @@ def push_edge_stmt(toks: ParseResults) -> T.List[pydot.core.Edge]:
 
     e = []
 
-    if isinstance(toks[0][0], pydot.core.Graph):
-        n_prev = FrozenDict(toks[0][0].obj_dict)
-    else:
-        n_prev = toks[0][0] + do_node_ports(toks[0])
+    def make_endpoint(
+        ep: T.Union[pydot.core.Common, T.List[T.Any], str],
+    ) -> T.Union[FrozenDict, str]:
+        if isinstance(ep, (list, tuple)) and len(ep) == 1:
+            # This is a hack for the Group()ed edge_point definition
+            ep = ep[0]
+        if isinstance(ep, pydot.core.Subgraph):
+            ep.obj_dict["show_keyword"] = False
+            return FrozenDict(ep.obj_dict)
+        if isinstance(ep, (list, tuple)):
+            return str(ep[0]) + do_node_ports(ep)
+        return str(ep)
 
-    if isinstance(toks[2][0], ParseResults):
-        n_next_list = [[n.get_name()] for n in toks[2][0]]
-        for n_next in list(n_next_list):
-            n_next_port = do_node_ports(n_next)
-            e.append(pydot.core.Edge(n_prev, n_next[0] + n_next_port, **attrs))
+    endpoints = [t for t in toks.as_list() if not isinstance(t, P_AttrList)]
 
-    elif isinstance(toks[2][0], pydot.core.Graph):
-        e.append(
-            pydot.core.Edge(n_prev, FrozenDict(toks[2][0].obj_dict), **attrs)
-        )
-
-    elif isinstance(toks[2][0], pydot.core.Node):
-        node = toks[2][0]
-
-        name_port: str
-        if node.get_port() is not None:
-            name_port = node.get_name() + ":" + node.get_port()  # type: ignore
-        else:
-            name_port = node.get_name()
-
-        e.append(pydot.core.Edge(n_prev, name_port, **attrs))
-
-    # if the target of this edge is the name of a node
-    elif isinstance(toks[2][0], str):
-        for n_next in list(tuple(toks)[2::2]):
-            if isinstance(n_next, P_AttrList) or not isinstance(
-                n_next[0], str
-            ):
-                continue
-
-            n_next_port = do_node_ports(n_next)
-            e.append(pydot.core.Edge(n_prev, n_next[0] + n_next_port, **attrs))
-
-            n_prev = n_next[0] + n_next_port  # type: ignore
-    else:
-        raise Exception(
-            f"Edge target {toks[2][0]} with type {type(toks[2][0])}"
-            " unsupported."
-        )
+    n_prev = make_endpoint(endpoints[0])
+    for endpoint in endpoints[1:]:
+        n_next = make_endpoint(endpoint)
+        e.append(pydot.core.Edge(n_prev, n_next, **attrs))
+        n_prev = n_next
 
     return e
 
@@ -466,8 +443,6 @@ class GraphParser:
 
     attr_stmt = Group(graph_ | node_ | edge_) + attr_list
 
-    edgeop = Literal("--") | Literal("->")
-
     stmt_list = Forward()
     graph_stmt = Group(
         lbrace.suppress()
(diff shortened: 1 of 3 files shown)


ferdnyc:
> This PR makes minimal changes to the parser

I say that that way because one thing has become clear while working on this: The parser is a _mess_ and really needs an overhaul. Currently any changes need to be minimal, to avoid destabilizing the entire thing and breaking it.

The issue is mostly with the parse actions. Here's a typical one:

```python
def push_subgraph_stmt(toks: ParseResults) -> pydot.core.Subgraph:
    g = pydot.core.Subgraph("")
    for e in toks:
        if len(e) == 3:
            e[2].set_name(e[1])
            if e[0] == "subgraph":
                e[2].obj_dict["show_keyword"] = True
            return e[2]  # type: ignore
        else:
            if e[0] == "subgraph":
                e[1].obj_dict["show_keyword"] = True
            return e[1]  # type: ignore

    return g
```

The assumptions built in there about the _exact_ structure of the `ParseResults` token list passed in as `toks` are MASSIVE, and the tiniest change _anywhere_ in the parser definition can have ripple effects that break that fragile code.

And it doesn't need to be that way. PyParsing [best practices](https://pyparsing-docs.readthedocs.io/en/latest/HowToUsePyparsing.html#usage-notes) include this note about results names:

> * Results names are recommended for retrieving tokens from complex expressions. It is much easier to access a token using its field name than using a positional index, especially if the expression contains optional elements. You can also shortcut the set_results_name call:
>
>     ```python
>     stats = ("AVE:" + real_num.set_results_name("average")
>              + "MIN:" + real_num.set_results_name("min")
>              + "MAX:" + real_num.set_results_name("max"))
>     ```
>     can more simply and cleanly be written as this:
>     ```python
>     stats = ("AVE:" + real_num("average")
>              + "MIN:" + real_num("min")
>              + "MAX:" + real_num("max"))
>     ```

That _absolutely_ applies. The `subgraph` definition in the parser is currently:

```python
subgraph = Group(subgraph_ + Optional(ID) + graph_stmt)
```

Which explains the parse action structure:

* The outer `for` loop is because `subgraph` is a `Group()`, which means it always returns its results as a list of `ParseResults` objects (though there will only ever be one object in this list).
    * If the (inner) list has three elements in it, then the `Optional(ID)` matched and the `pydot.core.Subgraph` object created by `graph_stmt`'s parse action is at index `2`. 
    * So the code checks whether `e[0]` is `"subgraph"` (even though it can't be anything else), and if it is sets the `Subgraph` object's `show_keyword` parameter to `True`.
    * If the list doesn't have 3 elements (then it will have 2, since only one of the 3 matching elements is optional), then the code again pointlessly checks whether `e[0]` is `"subgraph"`, before again updating the `Subgraph` object (now at position `e[1]`) so that its `show_keyword` parameter is `True`.

Now, there are two ironies here:

1. None of those checks need to be done! This is a parse action on the `subgraph` statement, which **only runs** when the parser has encountered a `subgraph_` statement -- meaning, the literal word `"subgraph"`. The very fact that this action is even running already indicates that `show_keyword` should be `True`.
2. `show_keyword` _defaults_ to `True` in `pydot.core.Subgraph`, so the work this action is doing is useless -- it's just reaffirming the existing value. Really, there should be code in the `graph_stmt` parse action that sets `show_keyword` to `False`, which this action would then override to correct. But there isn't any such code. (That's why I had to hack an `obj_dict["show_keyword"] = False` into my updated `push_edge_stmt` code, to avoid an edge like `a -- { b c };` coming out as:
    ```dot
    graph G {
      a -- subgraph {
        b;
        c;
      };
    }
    ```

    ...which _works_ (`dot` supports that), but isn't what the input said.

But, both of those ironies aside, if the `subgraph` definition instead looked like this:

```python
subgraph = subgraph_("kw"} + Optional(ID("sg_name")) + graph_stmt("sg")
```

...Then the parse action could look like this:

```python
def push_subgraph_stmt(toks: ParseResults) -> pydot.core.Subgraph:
    assert "sg" in toks
    g = toks["sg"]
    g.obj_dict["show_keyword"] = "kw" in toks
    if "sg_name" in toks:
        g.set_name(toks["sg_name"])
    return g
```

ferdnyc:
Oh, Ruff, you loveable curmudgeon.

ferdnyc:
> Really, there should be code in the `graph_stmt` parse action that sets `show_keyword` to `False`, which this action would then override to correct. But there isn't any such code. (That's why I had to hack an `obj_dict["show_keyword"] = False` into my updated `push_edge_stmt` code, to avoid an edge like `a -- { b c };` coming out as:
>     ```dot
>     graph G {
>       a -- subgraph {
>         b;
>         c;
>       };
>     }
>     ```
>       
>     ...which _works_ (`dot` supports that), but isn't what the input said.

So, I realized after writing this (and partly _because of_ writing this) that I'd actually broken things in the opposite direction. By forcing `show_keyword=False` in `push_edge_stmt`, I'd broken the ability to _actually write_ an edge like this:

```dot
a -- subgraph XY { b; c; };
```

That's supposed to be legal, but with my changes would've come out as:

```dot
a -- {
b;
c;
};
```

So I went back in, removed my `push_edge_stmt` hack, and changed `push_graph_stmt` to set `show_keyword=False` the way it always should've been doing. That way, the code in `push_subgraph_stmt` that sets it _back_ to `True` for explicit subgraphs is actually doing something useful, and now all subgraphs that should or shouldn't be showing the keyword all work the way they're supposed to.

A couple of new test functions in `test_parser.py` confirm that all is as it should be.

lkk7:
Thanks, I'll release this soon after some checks.

ferdnyc:
> Not only is that (a) not incredibly fragile and (b) more efficient, but it even captures and correctly applies the subgraph's name, when set.
> 
> (The current parser _does_ manage to capture the subgraph's name, so some other parse action must be handling that. Even though this parse action is the only _correct_ place to do it. But since some other code has to therefore contain an assumption about the structure of the `subgraph` `ParseResults`, that other code will almost certainly break if the `subgraph` definition and parse action were changed. That's what I mean about the fragility issue.)

Ignore what I said about name-setting, there... it's a testament to the utter unreadability of the current parser's parse action(s) code, with all of its dereferencing-ParseResults-tokens-using-random-seeming-numerical-indices nonsense, that I managed to not even notice the `set_name()` call buried in the existing parse action:

```python
def push_subgraph_stmt(toks: ParseResults) -> pydot.core.Subgraph:
    g = pydot.core.Subgraph("")
    for e in toks:
        if len(e) == 3:
            e[2].set_name(e[1])  #  <= (THIS ONE HERE)
            if e[0] == "subgraph":
                e[2].obj_dict["show_keyword"] = True
            return e[2]  # type: ignore
        else:
            if e[0] == "subgraph":
                e[1].obj_dict["show_keyword"] = True
            return e[1]  # type: ignore

    return g
```

ferdnyc:
@lkk7 No rush. Anything I can do to help things along?

lkk7:
@ferdnyc  Sure, if you prepare changelog entries, that would be helpful. If not, I'll do it when I come back.

And I'm wondering if, in the future, "add entry to changelog if applicable" should be a required step to tick off in all PRs.

ferdnyc:
@lkk7 In my experience that can be fraught with peril.

1. If every PR modifies the same ChangeLog file, they'll all conflict with each other which gets annoying fast. There are systems that can work around that (usually by integrating multiple separate "log-snippet" files into a single changelog at release) but that's a lot of complexity.

2. More relevantly, when merging code to the repo you never actually know for sure what release(s) it's going to be a part of. Unless releases are strictly linear, it's difficult to organize that stuff in advance/incrementally.

Have you tried out GitHub's release-creation tools? If you create a release using the web interface, it auto-populates the release notes with a list of every PR that's been merged since the branching point, which is usually a pretty good starting point. Most of the time, I find it's just a matter of _removing_ the internal housekeeping PRs from the list, and done.

PR #469, opened 2025-04-24
Fix a few type annotations

lkk7:
I'm waiting for the CI results, as I'm not sure if my `mypy` is showing problems that CI doesn't see, or if we should really push these fixes.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -568,7 +568,7 @@ class Common:
         self.obj_dict["parent_graph"] = parent_graph
 
     def get_parent_graph(self) -> Optional["Graph"]:
-        return self.obj_dict.get("parent_graph", None)  # type: ignore
+        return self.obj_dict.get("parent_graph", None)
 
     def get_top_graph_type(self, default: str = "graph") -> str:
         """Find the topmost parent graph type for the current object."""
@@ -1434,9 +1434,9 @@ class Graph(Common):
             edge_src_set, edge_dst_set = list(
                 zip(*[obj["points"] for obj in edge_obj_dicts])
             )
-            edge_src_set, edge_dst_set = set(edge_src_set), set(edge_dst_set)  # type: ignore
+            edge_src_set, edge_dst_set = set(edge_src_set), set(edge_dst_set)
         else:
-            edge_src_set, edge_dst_set = set(), set()  # type: ignore
+            edge_src_set, edge_dst_set = set(), set()
 
         node_obj_dicts = []
         for k in self.obj_dict["nodes"]:
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -84,7 +84,7 @@ class HTML(Token):
     """Parsing for HTML-like strings."""
 
     def __init__(self) -> None:
-        super().__init__()
+        super().__init__()  # type: ignore
 
     def parseImpl(
         self, instring: str, loc: int, do_actions: bool = True


src/pydot/dot_parser.py
```diff
@@ -84,7 +84,7 @@ class HTML(Token):
     """Parsing for HTML-like strings."""
 
     def __init__(self) -> None:
-        super().__init__()
+        super().__init__()  # type: ignore
```
ferdnyc, line 87: This is only necessary in Python 3.8, and in later versions `mypy --strict` will actually complain about it. That's why I removed it in #464, and changed the CI definition to do MyPy checks in Python 3.9 and 3.13, instead of 3.8 and 3.13. Writing our type annotations for an unsupported Python release, at the expense of current releases, feels like the wrong way to break on that one.

src/pydot/core.py
```diff
@@ -568,7 +568,7 @@ def set_parent_graph(self, parent_graph: Optional["Common"]) -> None:
         self.obj_dict["parent_graph"] = parent_graph
 
     def get_parent_graph(self) -> Optional["Graph"]:
-        return self.obj_dict.get("parent_graph", None)  # type: ignore
+        return self.obj_dict.get("parent_graph", None)
```
ferdnyc, line 571: I'm surprised this is OK without the ignore, because the _setter_ defines its `parent_graph` input as `Optional["Common"]`. So, returning (what's potentially been set to) an `Optional["Common"]` as an `Optional["Graph"]` feels like it shouldn't work. But, if MyPy is OK with it... `¯\_(ツ)_/¯`

src/pydot/core.py
```diff
@@ -1434,9 +1434,9 @@ def to_string(
             edge_src_set, edge_dst_set = list(
                 zip(*[obj["points"] for obj in edge_obj_dicts])
             )
-            edge_src_set, edge_dst_set = set(edge_src_set), set(edge_dst_set)  # type: ignore
+            edge_src_set, edge_dst_set = set(edge_src_set), set(edge_dst_set)
         else:
-            edge_src_set, edge_dst_set = set(), set()  # type: ignore
+            edge_src_set, edge_dst_set = set(), set()
```
ferdnyc, line 1439: These, OTOH, make total sense.

src/pydot/core.py
ferdnyc: Yeah, see, mypy complained about it:

```text
src/pydot/core.py:571: error: Returning Any from function declared to return "Optional[Graph]"  [no-any-return]
```

src/pydot/core.py
ferdnyc: Oh, no they don't, because the variables are initially created from a `zip()` so they're tuples.

Best solution for this is probably...

```python
if edge_obj_dicts:
    edge_srcs, edge_dsts = list(
        zip(*[obj["points"] for obj in edge_obj_dicts])
    )
    edge_src_set, edge_dst_set = set(edge_srcs), set(edge_dsts)
else:
    edge_src_set, edge_dst_set = set(), set()
```

ferdnyc: Actually, wait, never mind. I always meant to fix this.

There's _no reason_ to keep separate lists of the two edge-endpoint sides.

These sets are only used for `suppress_disconnected` mode, which only cares if a node has no connections. It doesn't MATTER which endpoint of the edge it's on, so the endpoints can all be kept in one big list instead of dealing with two separate.

I'll put together a quick PR to fix that the right way.

lkk7:
Yes, as expected, this was a `mypy` hallucination on my end. Although I checked, I may have been using an old system-wide `mypy' instead of the venv one due to some path conflicts.

ferdnyc:
@lkk7 That's OK, I found another typing issue related to complex edges — the endpoints of an Edge can (now) be `FrozenDict`, which was accounted for in the `.to_string()` method but _**not**_ in its constructors and getters. (Some `#type: ignore`s were hiding the issue.)

Working on fixing that now, with an `EdgeEndpoint` type alias that allows `str | int | float | FrozenDict`. That seems to be the true set of types we want to _actually_ permit in `Edge`s, now.

PR #470, opened 2025-04-25
Type edge endpoints, tighten getter typing, and improve suppress_disconnected and strict modes (with tests)

ferdnyc:
### Add two new type aliases related to `Edge` objects

- `EdgeDefinition` represents the input type for the `src` and `dst` arguments of `Edge.__init__`. It can be a `Node`, `Subgraph`, `Cluster`, `FrozenDict`, or `int`, `float`, or `str`
- `EdgeEndpoint` represents the _output_ type, which can be any of `FrozenDict`, `int`, `float`, or `str`. Any Pydot classes passed in have been dereferenced to their `.get_name()` value.

### Tighten up typing of setters, getters

By making methods like `get_strict()` and `get_simplify()` no longer return `Optional` values, but forcing a `bool` result, we can use those methods in `.to_string()` and other places without running afoul of the type system.

### Improve & test `strict`, `suppress_disconnected`

When suppressing disconnected nodes, we only care if a node is on _any_ endpoint of an edge — we don't care if it's the `src` or `dst` endpoint. So, instead of keeping separate `edge_src_set` and `edge_dst_set` to compare against, glom them all into a single `edge_ep_set` that the node can be checked against.

Add new `test_api` test functions for both `strict` and `suppress_disconnected`

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -6,6 +6,7 @@
 
 import copy
 import errno
+import itertools
 import logging
 import os
 import re
@@ -1431,12 +1432,13 @@ class Graph(Common):
             edge_obj_dicts.extend(self.obj_dict["edges"][k])
 
         if edge_obj_dicts:
-            edge_src_set, edge_dst_set = list(
-                zip(*[obj["points"] for obj in edge_obj_dicts])
+            edge_ep_set = set(
+                itertools.chain.from_iterable(
+                    obj["points"] for obj in edge_obj_dicts
+                )
             )
-            edge_src_set, edge_dst_set = set(edge_src_set), set(edge_dst_set)  # type: ignore
         else:
-            edge_src_set, edge_dst_set = set(), set()  # type: ignore
+            edge_ep_set = set()
 
         node_obj_dicts = []
         for k in self.obj_dict["nodes"]:
@@ -1452,16 +1454,13 @@ class Graph(Common):
         ]
         obj_list.sort(key=lambda x: x[0])
 
+        skip_disconnected = self.obj_dict.get("suppress_disconnected", False)
         for idx, obj in obj_list:
             if obj["type"] == "node":
                 node = Node(obj_dict=obj)
 
-                if self.obj_dict.get("suppress_disconnected", False):
-                    if (
-                        node.get_name() not in edge_src_set
-                        and node.get_name() not in edge_dst_set
-                    ):
-                        continue
+                if skip_disconnected and node.get_name() not in edge_ep_set:
+                    continue
 
                 node_str = node.to_string(
                     indent=indent, indent_level=indent_level + 1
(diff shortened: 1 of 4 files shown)


ferdnyc:
Now let's see if Ruff spanks me.

src/pydot/core.py
```diff
@@ -1452,16 +1454,13 @@ def to_string(
...
                 node = Node(obj_dict=obj)
 
-                if self.obj_dict.get("suppress_disconnected", False):
-                    if (
-                        node.get_name() not in edge_src_set
-                        and node.get_name() not in edge_dst_set
-                    ):
-                        continue
+                if skip_disconnected and node.get_name() not in edge_ep_set:
+                    continue
```
ferdnyc, line 1463: Wait, do we _not have_ a test that exercises `suppress_disconnected`?!? (And actually has disconnected nodes to suppress.)

Huh! We **don't**. In fact, we [don't have any coverage](https://htmlpreview.github.io/?https://raw.githubusercontent.com/pydot/pydot/python-coverage-comment-action-data/htmlcov/z_288b311da95ac7b6_core_py.html#t1459) for `suppress_disconnected` mode **AT ALL**.

Suppose I should fix that.

lkk7: Will you create the tests in this PR?

ferdnyc:
OK, the new commits solve the `Edge` typing issue I mentioned in #469, tighten up the typing on a bunch of getters and setters by eliminating the `Optional` form of their input/return values, and goose the coverage of `Graph.to_string()` by testing _both_ `strict` and `suppress_disconnected`, neither of which was previously covered.

src/pydot/core.py
```diff
@@ -606,17 +607,20 @@ def get(self, name: str) -> Any:
         """
         return self.obj_dict["attributes"].get(name, None)
 
-    def get_attributes(self) -> Optional[AttributeDict]:
+    def get_attributes(self) -> AttributeDict:
         """Get attributes of the object"""
-        return self.obj_dict["attributes"]  # type: ignore
+        return cast(AttributeDict, self.obj_dict.get("attributes", {}))
```
lkk7, line 612: It really hurts my inner OCD to know that in Python sometimes you have to call a function to satisfy a static checker. I know it's probably not a big deal, it's optimized and all, but still.

lkk7 (approved):

src/pydot/core.py
ferdnyc: Whoops, no. It was 3.11, not 3.10, that [sped up basically everying](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-faster-cpython), including function calls.

Before 3.11, even calling something like `a.__add__(b)` or `operator.add(a, b)` was 2-10× slower than just `a + b`.

Issue #474, opened 2025-05-06
ImportError: cannot import name 'DelimitedList' from 'pyparsing' (pyparsing dependency version not up to date?)


juusomer:
If you have pyparsing 3.0.9 installed, importing from `pydot.dot_parser` will fail.

```
% pip install pyparsing==3.0.9 && pip install pydot
% python -c 'from pydot.dot_parser import *'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/juuso.merilainen/work/projects/.venv/lib/python3.11/site-packages/pydot/dot_parser.py", line 18, in <module>
    from pyparsing import (
ImportError: cannot import name 'DelimitedList' from 'pyparsing' (.venv/lib/python3.11/site-packages/pyparsing/__init__.py)
```

The pyparsing dependency allows 3.0.9 (https://github.com/pydot/pydot/blob/main/pyproject.toml#L19), but this does not seem to have the `DelimitedList` class imported here https://github.com/pydot/pydot/blob/c37e44bfdadb31e91b4e3a96e0c832af65d60c12/src/pydot/dot_parser.py#L21 Based on my understanding it was only introduced in 3.1.0 (https://pyparsing-docs.readthedocs.io/en/latest/whats_new_in_3_1.html#id4).

lkk7:
Looks like you're right. The import that fails was introduced by a recent parsing renovation.
I'll prepare a PR soon to change the minimum version to `3.1.0` – that seems to be enough. Thanks!

juusomer:
Thanks! I was unsure whether simply bumping the dependency version would suffice, but made a PR with that change. If there are some additional things to check or some practices I should follow, do let me know.

PR #477, opened 2025-05-12
Update pyparsing dependency

juusomer:
This ensures that `DelimitedList` imported here https://github.com/pydot/pydot/blob/c37e44bfdadb31e91b4e3a96e0c832af65d60c12/src/pydot/dot_parser.py#L21 is available in `pyparsing`, as this was only introduced in 3.1.0 (https://pyparsing-docs.readthedocs.io/en/latest/whats_new_in_3_1.html#id4).

See issue https://github.com/pydot/pydot/issues/474

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ license-files = [
 ]
 requires-python = ">= 3.8"
 dependencies = [
-  'pyparsing>=3.0.9'
+  'pyparsing>=3.1.0'
 ]
 authors = [
   {name = "Ero Carrera", email = "ero.carrera@gmail.com"},


lkk7 (approved):

lkk7:
Thanks!

ferdnyc:
Yeah, sorry, I had no idea `DelimitedList` was so new-ish. It _is_ listed in [What's New in Pyparsing 3.1.0](https://pyparsing-docs.readthedocs.io/en/latest/whats_new_in_3_1.html) — it actually replaced a function, `delimited_list()`, from previous versions — but there's no indication of its recency if you're just reading the docs themselves.

Maybe I'll get on PMcG's case about adding "since"-type annotations to the API docs, at least for stuff introduced since 3.0.0.

Issue #473, opened 2025-05-04
Python 3.8 installs fine but has build issues


lkk7:
I got ahead of myself too quickly in: https://github.com/pydot/pydot/commit/8fc4016640fc4d0fa50f701d57694428184c1d3c

I tested it locally with tox and it worked on 3.8.
But I forgot about the important fact that the CI not only runs the 3.8 tests on tox, but that the *tox itself* is on 3.8. I'm not sure, but this might be the difference that makes the current CI fail.

In short, I fixed the deprecation warnings, but that broke the manual build for 3.8.
3.8 users can install pydot with no problems, but they probably can't build it from source.

Since 3.8 is deprecated, EOL, and [used to download pydot very rarely](https://pypistats.org/packages/pydot), I wonder if we should care. There's a good chance no one will ever use 3.8 to *build* pydot.
So I'm wondering how to fix this in our CI. Or maybe go back to the deprecated methods (we have time until Feb 2026 to change that).

Anyway, that was stupid to do right before a release!

@ferdnyc

ferdnyc:
@lkk7 

I just tested with a `python3.8`-from-the-jump environment.

The build issue is that the `pyproject.toml` license syntax is out of date. This diff makes it buildable with 3.8:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 327c129..7b59eec 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,10 +10,7 @@ build-backend = "setuptools.build_meta"
 name = 'pydot'
 description = "Python interface to Graphviz's Dot"
 readme = "README.md"
-license = "MIT"
-license-files = [
-   'LICENSES/*',
-]
+license = { file = 'LICENSES/*' }
 requires-python = ">= 3.8"
 dependencies = [
   'pyparsing>=3.0.9'

```

ferdnyc:
(You can't list both `file` and `text` -- that's the other subkey option -- or it errors out. And IIRC using both `license-files = ["foo"]` and `license = "bar"` is interpreted the same as `license = {file = "foo", text = "bar" }`

ferdnyc:
To apply a _well-known license_ by **name** (like "MIT" or "GPL-3.0-or-later"), IIRC the advice was "Do it through the Trove classifiers instead". There are classifiers for every common OSS license.

ferdnyc:
[...]

PR #478, opened 2025-05-13
Add __future__.annotations imports

ferdnyc:
Conversation in #473 led to the discovery that in Python 3.7-3.9, forward-referencing annotations were only supported in the presence of this import. Add it to make type-annotating code easier in the future.

This then required adjusting _many_ of our types, with Ruff suddenly insisting on `foo | None` syntax instead of `Optional[foo]`, and on use of real container types in annotations rather than the `typing` import representations (e.g. `list[str]` not `typing.List[str]`). I think the appearance of `__future__.annotations` imports pushed it to upgrade the implicit minimum version it uses for syntax checks.

A second commit reduces the `typing` imports in `dot_parser` to only `Any`, since no other `typing`-namespace symbols are required anymore.

(`core` still uses `Sequence` and `cast`, `classes` still uses `Dict`, and both still use `Union`. The latter two are because of the backwards-compatible type aliases: `AttributeDict = Dict[str, Any]` can't be rewritten as `AttributeDict = dict[str, Any]`, for example, even though that's supported _as an annotation_ now.)

### DIFF
--- a/src/pydot/classes.py
+++ b/src/pydot/classes.py
@@ -4,8 +4,10 @@
 
 """Frozen dictionaries."""
 
+from __future__ import annotations
+
 import copy
-from typing import Any, Dict, Tuple, Union
+from typing import Any, Dict, Union
 
 
 class FrozenDict(dict):  # type: ignore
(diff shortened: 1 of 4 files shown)


lkk7:
Nice! I much prefer using the new "real" types instead of their substitutes from `typing`.

lkk7 (approved):

ferdnyc:
@lkk7 Same. And I've been jonesing for `|` syntax to replace `Union` for a _while_ now. Especially since it's already a major part of the parser syntax, so the symmetry is pleasing.

PR #480, opened 2025-05-15
Fix 'strict' handling in parser

### LINKED ISSUE #479: pydot looses strictness of graphs for `graph_from_file` and `graph_from_dot_data`
with a  strict graph, for e.g. the following as a string or a file (named for example `dotfile`) : 
```dot
strict graph G {}
```
when doing:
```python
pydot.graph_from_file("dotfile")[0].to_string()
```
or
```python
pydot.graph_from_dot_data("strict graph G {}")[0].to_string()
```
for both invocations, you get:
```dot
'graph G {\n}\n'
```
but one should expect the strictness of the graph to be preserved like so:
```dot
'strict graph G {\n}\n'
```
as is done when using `pygraphviz.AGraph(string="strict graph G {}")` or `pygraphviz.AGraph(filename="dotfile")`

I think the strictness of the graph is being ignored here.

ferdnyc:
It was reported that the parser wasn't capturing the 'strict' keyword
when parsing dot data. To ensure that it does, rewrite the top-level
`GraphParser.parser` definition to use results names, and rewrite
`push_top_graph_stmt()` in a streamlined way that makes use of them.

A second commit adds some tests.

Fixes #479

### DIFF
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -120,7 +120,7 @@ def push_top_graph_stmt(toks: ParseResults) -> list[pydot.core.Dot]:
         assert isinstance(result, ParseResults)
         gtype = result.gtype
         strict = "strict" in result
-        id_ = result.id or ""
+        id_ = str(result.id)
 
         g = pydot.core.Dot(id_, graph_type=gtype, strict=strict)
         g.set_parent_graph(g)
--- a/src/pydot/dot_parser.py
+++ b/src/pydot/dot_parser.py
@@ -113,57 +113,21 @@ class HTML(Token):
         )
 
 
-def push_top_graph_stmt(
-    s: str, loc: int, toks: ParseResults
-) -> list[pydot.core.Dot] | pydot.core.Dot:
-    attrs = {}
+def push_top_graph_stmt(toks: ParseResults) -> list[pydot.core.Dot]:
     top_graphs = []
-    g: pydot.core.Dot = None  # type: ignore
 
-    for element in toks:
-        if (
-            isinstance(element, (ParseResults, tuple, list))
-            and len(element) == 1
-            and isinstance(element[0], str)
-        ):
-            element = element[0]
+    for result in toks.graphs:
+        assert isinstance(result, ParseResults)
+        gtype = result.gtype
+        strict = "strict" in result
+        id_ = result.id or ""
 
-        if element == "strict":
-            attrs["strict"] = True
-
-        elif element in ["graph", "digraph"]:
-            attrs = {}
-
-            g = pydot.core.Dot(graph_type=element, **attrs)
-            attrs["type"] = element
-
-            top_graphs.append(g)
-
-        elif isinstance(element, str):
-            g.set_name(element)
-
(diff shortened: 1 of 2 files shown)


ferdnyc:
Oh, BTW: as part of the streamlining, `push_top_graph_stmt` now always returns a _list_ of one or more `pydot.core.Dot` objects, it no longer strips out a single graph object and returns it directly. We were re-wrapping the results inside a `list()` in `parse_dot_data()` anyway, so it was silly to ever make it temporarily not-a-list.

src/pydot/dot_parser.py
```diff
@@ -113,57 +113,21 @@ def parseImpl(
...
-
-        elif element in ["graph", "digraph"]:
-            attrs = {}
-
-            g = pydot.core.Dot(graph_type=element, **attrs)
-            attrs["type"] = element
 
-            top_graphs.append(g)
+    for result in toks.graphs:
+        assert isinstance(result, ParseResults)
```
lkk7, line 120: Is this `assert` made for satisfying `mypy`? Or left from debugging

ferdnyc: For appeasing our type overlords, for sure.

By default everything extracted from a `toks` argument can be of type:  `Any | list[Any] | Unknown | ParseResults | Literal['']`. Which causes trouble using it anywhere that won't accept all of those types.

src/pydot/dot_parser.py
```diff
@@ -113,57 +113,21 @@ def parseImpl(
...
-        elif isinstance(element, (ParseResults, list)):
-            add_elements(g, element)
-
-        else:
-            raise ValueError(f"Unknown element statement: {element}")
-
-    for g in top_graphs:
+        g = pydot.core.Dot(id_, graph_type=gtype, strict=strict)
+        g.set_parent_graph(g)
+        if isinstance(result.contents, ParseResults):
```
ferdnyc, line 127: (Same reason this check is here, though in this case `result.contents` actually could come back as `''` if it's parsing an empty `{ }`.)

src/pydot/dot_parser.py
```diff
@@ -113,57 +113,21 @@ def parseImpl(
...
-
-            g = pydot.core.Dot(graph_type=element, **attrs)
-            attrs["type"] = element
 
-            top_graphs.append(g)
+    for result in toks.graphs:
+        assert isinstance(result, ParseResults)
+        gtype = result.gtype
+        strict = "strict" in result
+        id_ = result.id or ""
```
ferdnyc, line 123: This `or` here, OTOH, I'm now realizing is dumb. Named results in `ParseResults` objects _default_ to returning `''` if they haven't captured any data. They'll never be `None`, as I was apparently assuming might happen when I wrote this.

I mean, it's harmless as such things go; if `result.id` is `''` the `or` will set `id_` to the same thing it would've been set to anyway. `¯\_(ツ)_/¯`

ferdnyc, line 123: Eh, still...

```suggestion
        id_ = str(result.id)
```

ferdnyc: Fixed it anyway.

PR #482, opened 2025-05-16
(Nearly) complete dot_parser coverage

ferdnyc:
When looking into #479, I [had the realization](https://github.com/pydot/pydot/issues/479#issuecomment-2885152918) that if we had complete coverage in the parser code, we'd have caught that the parser was never applying the 'strict' keyword to graphs. That line had no test coverage (even though it should've), and even if we'd attempted to _directly_ add coverage for that line, we'd never have succeeded in reaching it. Which would've been a red flag that there was something wrong with the code.

So, this PR updates the remainder of the parser code to not only make a lot of the parsing cleaner (thanks to results names and a new helper function), but eliminate _tons_ of dead code that was never being called (and was often completely unreachable).

Direct testing for the `__repr__`s on `P_AttrList` and `DefaultStatement` is also added, to ensure coverage of that code.

The outcome is coverage for `dot_parser.py` jumping up to 93%, and except for the 5 lines of exception handling in `parse_dot_data()`, all of the uncovered lines are in `push_top_graph_stmt` (so they should be handled by #480). Once both of these PRs are merged we should have total coverage of all the parsing code, meaning we can be fairly certain that all of its features are operating properly.

Along the way I also made a few other adjustments:

* Renamed the `attr_stmt` parser element to `default_stmt` like it always should've been.
* Eliminated `port` parsing _entirely_. See the "improve node parsing" for a full explanation of that one
* Got rid of all the misguided defaults handling in `add_elements`, again see the "Streamline add_elements" commit message for details


ferdnyc:
Oh, yes, see also the "Clean up update_parent_graph_hierarchy" commit message for some details on why we may want to look deeper into that function and the reasons it exists at all.

lkk7:
Ruff will always catch by surprise.

PR #486, opened 2025-06-01
Long names fix

### LINKED ISSUE #483: Long node names get mangled by `create_dot`
_Originally reported by @dougthor42 in https://github.com/networkx/networkx/issues/7648._

When working with long node names (~ >130 characters), `create_dot` mangles the name by adding line breaks. This makes it hard to round-trip `pydot.Graph` objects through the dot command and read them back in again (which is what `nx.nx_pydot.pydot_layout` does).

A minimum reproduction is
```python
>>> import networkx as nx
>>> import pydot
>>> import locale
>>> G = nx.Graph()
>>> G.add_node("=10chars= " * 14)
>>> P = nx.nx_pydot.to_pydot(G)
>>> bytes = P.create_dot()
>>> D = str(bytes, encoding=locale.getpreferredencoding())
>>> Q = pydot.graph_from_dot_data(D)[0]
>>> print(Q.get_nodes()[-1].get_name())
"=10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =\
10chars= "
>>> print(pydot.quote_id_if_necessary(str(list(G.nodes())[0])))
"=10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= =10chars= "
```
The two names aren't equal, and it would be nice if they were.

lkk7:
Fixes #483  (?)

This is just a draft to test what's happening.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -569,7 +569,7 @@ class Common:
         self.obj_dict["parent_graph"] = parent_graph
 
     def get_parent_graph(self) -> Graph | None:
-        return self.obj_dict.get("parent_graph", None)  # type: ignore
+        return self.obj_dict.get("parent_graph", None)
 
     def get_top_graph_type(self, default: str = "graph") -> str:
         """Find the topmost parent graph type for the current object."""
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -1184,8 +1184,10 @@ class Graph(Common):
             )
         else:
             for key, node_list in self.obj_dict["nodes"].items():
-                if "\\\n" in key:
-                    normalized_key = key.replace("\\\n", "")
+                if "\\\n" or "\\\r\n" in key:
+                    normalized_key = key.replace("\\\n", "").replace(
+                        "\\\r\n", ""
+                    )
                     if normalized_key == name:
                         match.extend(
                             Node(obj_dict=obj_dict) for obj_dict in node_list
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -1175,7 +1175,6 @@ class Graph(Common):
         An empty list is returned otherwise.
         """
         match = []
-
         if name in self.obj_dict["nodes"]:
             match.extend(
                 [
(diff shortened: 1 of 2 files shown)


lkk7:
Now that I think of it, it still might not satisfy everyone because the name of the node we'll get by `get_node()` (split into two lines) does not equal the name we used to search for it (unsplit). But that might be acceptable.

Peiffap (commented):
Thanks for taking a look so quickly!

I'm trying to wrap my head around the implications of this (without being super familiar with the surrounding code...). Iiuc, the only situation where you could break someone's workflow is if they have two nodes with "long" names that are the same, except for at least one linebreak, right? I'm... not sure how common that would be. I might be missing something though!

lkk7:
> the only situation where you could break someone's workflow is if they have two nodes with "long" names that are the same, except for at least one linebreak

Yeah, that's the only thing I could think of.
But I'm willing to trade the risk of that for fixing this bug. Because people often make abominable long node names, but I don't think they often create multiple nodes which are the same except for a line break.
I'll try to check this evening how can this be broken manually.

Peiffap:
I think because you wrapped it in an `if else` that searches for exact matches first, this PR should still be a strict improvement: from failing for all long node names to now only failing when there are multiple long node names that are the same after removing all line breaks. Much less likely to happen, imo. I also can't think of any other situation that would break, but it's worth thinking about some more to ensure you don't get flooded with bug reports when this releases!

lkk7:
Yeah, there is a possibility for breakage in`Graph.add_node()` (which is also used internally), because it uses the `get_node()` we want to fix. You have to be intentional and add functionally the same node two times.

```python
import pydot

dot_string = """
strict graph {
"node\\
123456";
"node123456";
}
"""

graphs = pydot.graph_from_dot_data(dot_string) # breaks
graph = graphs[0]
```

That's the same thing as adding "abc" two times, but in a weird broken way. Still, maybe it might happen to someone someday.
I could explore adding a small workaround to `add_node`. Or maybe this scenario shouldn't even be considered.

Is this fix urgent for you?

PR #490, opened 2025-06-13
Parser: Support +-concatenated quoted strings

### LINKED ISSUE #488: Parser: Missing support for +-concatenated quoted strings
The GraphViz language documentation says, in the ["Comments and optional formatting" section](https://graphviz.org/doc/info/lang.html#comments-and-optional-formatting)):

> In addition, double-quoted strings can be concatenated using a '+' operator.

Our parser does not currently have any support for that syntax.

_Adapted from a comment originally posted by @ferdnyc in [#483](https://github.com/pydot/pydot/issues/483#issuecomment-2961183272)_

(Granted, I've never seen anyone actually _use_ that syntax, but I've tried it and it does work in `dot`, so we should probably support it as well.)

ferdnyc:
Anywhere a double-quoted string can appear in the `dot` code passed to the parser, it's now possible to instead provide `"multiple " + "concatenated " + "strings"`, which will be parsed as `"multiple concatenated strings"`.

`+`-concatenation is fully compatible with backslash continuations, and can appear in attribute values as well. This will create an orange node named "apple":

```dot
graph G {
    "app" +
    "le" [color="o" + "r" + "a\
n" + "g"
    + "e"];
}
```

Both a `test_plus_concatenation()` test function in `test_parser.py`, and a new test input file `test/graphs/b488.dot`, are added in the second commit.

Fixes #488

### DIFF
--- a/test/graphs/b488.dot
+++ b/test/graphs/b488.dot
@@ -1,0 +1,15 @@
+
+digraph G {
+    "concatenated" + "name" [color="bl" + "ue"];
+    "concatenated with " + "ports" [color="or\
+an" + "ge"];
+    "concatenated" + " with ports":p45:sw -> "\
+concatenated\
+name":ne [penwidth=5, arrows="b" + "o" + "t" + "h"];
+    "con" + "catenated" [label="this is a long" +
+    "long label" + " that just goes on \
+and on and on"];
+    "con" +
+    "catenated" -> "back\
+slashed" [shape=square];
+}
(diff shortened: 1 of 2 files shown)


ferdnyc:
You get me every time, Ruff. :stuck_out_tongue_closed_eyes:

lkk7 (approved):
Thanks!

test/graphs/b488.dot
```diff
@@ -0,0 +1,15 @@
+
```
lkk7, line 1: I forgot to ask – why `b488.dot` for the graph name in particular? 😄

ferdnyc: My issue for the missing syntax support was #488.

PR #492, opened 2025-06-17
Tighten up our API with keyword-only arguments

ferdnyc:
To tighten up the way the API is used, add an `*` to the argument
list of appropriate methods, to signify that the arguments beyond
that point are keyword-only.

Specifically...

- in `to_string` methods, `indent` is keyword-only (as is the
  internal `indent_level`)

- in `del_` methods, the optional `index` is keyword-only

- `quote_id_if_necessary`'s `unquoted_keywords` is keyword-only

- in all constructors except `Common.__init__`, `obj_dict` is
  keyword-only, and moved to the end of the keyword argument list.
  (It's not kw-only in `Common`, because it's the ONLY argument and
  the other constructors call it as `super().__init__(obj_dict)`.)

- Constructor arguments like `strict`, `suppress_disconnected`, etc.
  for `Graph` types are keyword-only.

- `graph_type` is explicitly _not_ kw-only, meaning a graph
  could be constructed as `pydot.Graph("G", "digraph")` which seems
  perfectly fine.

- `Dot.create` used to be called internally by `Dot.write` as
  `Dot.create(prog, format, encoding=)` which didn't seem like a
  very comfortable API for outside users, so now the `create` method
  takes `format` as the only positional argument, and the call is
  changed to `Dot.create(format, prog=, encoding=)`

- `Dot.write` itself takes only the `path` as a positional argument,
  the rest are keyword-only. While it might be nice to allow something
  like `Dot.write('/some/path', 'pdf')`, since the format is the
  _first_ positional argument of `Dot.create()` that's potentially
  confusing. (Ideally maybe `Dot.write()` should be able to infer the
  format based on the filename.)

### Rationale

It used to be possible, at least in theory, to make calls like these:

```python
g = pydot.Dot("G", None, "digraph", True, False, True)

g.to_string(3)           # (indent=3)
g.to_string("    ")      # (indent="    ") 
g.del_edge("a", "b", 3)  # (index=3)

pydot.quote_id_if_necessary(some_id, ("node", "edge", "graph"))

d = pydot.Dot("", g.obj_dict)
d.write('/tmp/out.svg', "dot", "svg", "utf-8")
d.create("neato", "svg", "latin1")
```

...all of which seem inadvisable and best avoided.

We're pretty good about keyword args in our own code. (I only had to change _one_ call to conform with this, an ancient `create()` call in `Dot.write` that used a weird set of positional arguments.) But, still, better that it's enforced by the method definitions. So, create some enforcement.

With this change, it's made explicit that the correct way to make the calls above is:

```python
# (Either of these is fine)
g = pydot.Graph("G", "digraph", strict=True, simplify=True)
g = pydot.Graph("G", graph_type="digraph", strict=True, simplify=True)

g.to_string(indent=3)
g.to_string(indent="    ")
g.del_edge("a", "b", index=3)

pydot.quote_id_if_necessary(
    some_id, unquoted_keywords=("node", "edge", "graph")
)
d = pydot.Dot(obj_dict=g.obj_dict)
d.write('/tmp/out.svg', format="svg", prog="dot", encoding="utf-8")
d.create("svg", prog="neato", encoding="latin1")
```

### Other changes

A second commit tightens the type for `.to_string(indent)` from `Any` to `str | int | float`.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -623,7 +623,7 @@ class Common:
         return int(seq)
 
     @staticmethod
-    def get_indent(indent: Any, indent_level: int) -> str:
+    def get_indent(indent: str | int | float, indent_level: int) -> str:
         if isinstance(indent, (int, float)):
             indent_str = " " * int(indent)
         else:
@@ -724,7 +724,9 @@ class Node(Common):
 
         self.obj_dict["attributes"]["style"] = ",".join(styles)
 
-    def to_string(self, indent: Any = "", indent_level: int = 1) -> str:
+    def to_string(
+        self, indent: str | int | float = "", indent_level: int = 1
+    ) -> str:
         """Return string representation of node in DOT language."""
         indent_str = self.get_indent(indent, indent_level)
 
@@ -896,7 +898,9 @@ class Edge(Common):
 
         return quote_id_if_necessary(node_ref)
 
-    def to_string(self, indent: Any = "", indent_level: int = 1) -> str:
+    def to_string(
+        self, indent: str | int | float = "", indent_level: int = 1
+    ) -> str:
         """Return string representation of edge in DOT language."""
         src = self.parse_node_ref(self.get_source())
         dst = self.parse_node_ref(self.get_destination())
@@ -1406,7 +1410,10 @@ class Graph(Common):
                 Graph(obj_dict=obj).set_parent_graph(parent_graph)
 
     def to_string(
-        self, indent: Any = "", indent_level: int = 0, inline: bool = False
+        self,
+        indent: str | int | float = "",
+        indent_level: int = 0,
+        inline: bool = False,
     ) -> str:
         """Return string representation of graph in DOT language.
 
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -346,7 +346,7 @@ def id_needs_quotes(s: str) -> bool:
 
 
 def quote_id_if_necessary(
-    s: str, unquoted_keywords: Sequence[str] | None = None
+    s: str, *, unquoted_keywords: Sequence[str] | None = None
 ) -> str:
     """Enclose identifier in quotes, if needed."""
     unquoted = [
@@ -673,6 +673,7 @@ class Node(Common):
     def __init__(
         self,
         name: str = "",
+        *,
         obj_dict: AttributeDict | None = None,
         **attrs: Any,
     ) -> None:
@@ -725,7 +726,7 @@ class Node(Common):
         self.obj_dict["attributes"]["style"] = ",".join(styles)
 
     def to_string(
-        self, indent: str | int | float = "", indent_level: int = 1
+        self, *, indent: str | int | float = "", indent_level: int = 1
     ) -> str:
         """Return string representation of node in DOT language."""
         indent_str = self.get_indent(indent, indent_level)
@@ -780,6 +781,7 @@ class Edge(Common):
         self,
         src: EdgeDefinition | Sequence[EdgeDefinition] = "",
         dst: EdgeDefinition = "",
+        *,
         obj_dict: AttributeDict | None = None,
         **attrs: Any,
     ) -> None:
@@ -899,7 +901,7 @@ class Edge(Common):
         return quote_id_if_necessary(node_ref)
 
     def to_string(
-        self, indent: str | int | float = "", indent_level: int = 1
+        self, *, indent: str | int | float = "", indent_level: int = 1
     ) -> str:
         """Return string representation of edge in DOT language."""
         src = self.parse_node_ref(self.get_source())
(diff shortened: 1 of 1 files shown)


ferdnyc:
This is technically an API change.

In fact, it's a fairly significant one _if_ you happen to be abusing the API with things like `pydot.Graph("name", None, "graph", True, True, False)`, because it will break those calls.

I think it's still a good idea, but I guess it should be targeted for Pydot 5.0.

lkk7 (approved):
This will probably break a lot of old code, but it's a cool change 😄 Definitely for 5.0! (BTW, 5.0? It flew by so quickly)
Maybe I'll scan some more well-known repos, like networkX, in my free time to see how this would affect them. Although good luck finding them there: https://github.com/pydot/pydot/network/dependents

ferdnyc:
Ideally I'd get un-lazy and add some tests before this gets merged, anyway. So, no rush.

ferdnyc:
In addition to the tests I've so far still been too lazy to write, I wonder if there's enough old, breakable code out there to make it worth providing a transitional wrapper layer for this?

Without too much work, I could add a module `src/pydot/legacy.py` that includes wrapper classes for all of the `core.py` ones, but using the old signatures. For example,

```python
# src/pydot/legacy.py
import pydot.core

class Node(pydot.core.Node):

   def __init__(
        self,
        name: str = "",
        obj_dict: AttributeDict | None = None,
        **attrs: Any,
    ) -> None:
    super().__init__(name, obj_dict=obj_dict, **attrs)


class Graph(pydot.core.Graph):

    def __init__(
        self,
        graph_name: str = "G",
        obj_dict: AttributeDict | None = None,
        graph_type: str = "digraph",
        strict: bool = False,
        suppress_disconnected: bool = False,
        simplify: bool = False,
        **attrs: Any,
    ) -> None:
        super().__init__(
            graph_name,
            graph_type,
            obj_dict=obj_dict,
            strict=strict,
            suppress_disconnected=suppress_disconnected,
            simplify=simplify,
            **attrs)

   def del_node(self, name: str | Node, index: int | None = None) -> bool:
       return super().del_node(name, index=index)

   # etc...
```

Might need some `@override` decorators to appease mypy, but basically that's the gist of it. Inheritance means that docstrings and unchanged methods will all be copied from the parent class.

Then, someone upgrading who has calls that violate the new API has the choice between _either_ fixing their calls, or changing their imports to pull in the `pydot.legacy` versions of the classes instead, as a stopgap. Because in a potentially large-ish codebase, fixing every pydot call could be a lot more work than this:

```diff
- from pydot import Graph, Node, Edge
+ from pydot.legacy import Graph, Node, Edge
```

Or, if I also import all of the _unchanged_ symbols into `pydot.legacy`, perhaps even this:

```diff
- import pydot
+ import pydot.legacy as pydot
```

Obviously, `pydot.legacy` would have to be born deprecated, and explicitly going away within one or two major releases at most.

...Ooh, one potential wrinkle: All of the methods that _create_ objects may also have to be overridden, since (for example) `Graph.get_node_list` is defined like this:

```python
    def get_node_list(self) -> list[Node]:
        node_objs: list[Node] = []
        for node in self.obj_dict["nodes"]:
            obj_dict_list = self.obj_dict["nodes"][node]
            node_objs.extend([Node(obj_dict=obj_d) for obj_d in obj_dict_list])
        return node_objs
```

...and for `pydot.legacy.Graph` we want it to be creating and returning `pydot.legacy.Node` objects, not `pydot.core.Node` objects.

(This is probably a sign that e.g. `pydot.core.Graph` should have methods like `new_node()`, `new_edge()`, etc. that it uses to create objects, instead of calling the constructors directly in its `get_*` methods (or any others). That way, the `new_*` methods can be overridden in the `legacy` subclass, as well as any other derived classes that we or anyone else may have, to tweak internal object creation without needing to make any other changes.)

lkk7:
I'd like to answer at length, but I agree with everything, so...
Yeah, it's a bit of work, and then there's the whole deprecation process. Need to make sure that we have clear deprecation warnings, and are prepared for annoyed users 😄 But this is one of the easiest migrations I've seen:
```diff
- import pydot
+ import pydot.legacy as pydot
```
So it's fine for me.

ferdnyc:
> But this is one of the easiest migrations I've seen:
> 
> ```diff
> - import pydot
> + import pydot.legacy as pydot
> ```

_If_ it works. I still have to puzzle through all of the implications of that, beyond the Graph class's uses of other class constructors.

Offhand, the entirety of `dot_parser` would also require some sort of "legacy mode switch" -- but I think that should be doable, especially now that `pydot.dot_parser` is no longer imported into `pydot.core`, but only imported on demand if `graph_from_dot_data` is called. That should make it easy for `pydot.legacy.graph_from_dot_data` to import a `dot_parser` configured to use `pydot.legacy` classes instead of `pydot.core` classes. (I hope.)

ferdnyc:
Wait... I'm an _idiot_.

Nothing about `dot_parser` has to change for `pydot.legacy` — the same extraction of `obj_dict`s and discarding of objects that makes subclassing a pain will actually _save_ us, there. `dot_parser` can go right on creating its graphs using `pydot.core.*` objects, and all `pydot.legacy.graph_from_dot_data` has to be is this:

```python
def graph_from_dot_data(s: str) -> list[pydot.legacy.Dot] | None:
    graphs = pydot.core.graph_from_dot_data(s)
    if graphs is not None:
        return [
            pydot.legacy.Dot(obj_dict=g.obj_dict)
            for g in graphs
        ]
    return graphs
```

Done.

PR #495, opened 2025-07-26
Make graph object classes subclassable

ferdnyc:
Because of the way pydot handles object data (extracting the `obj_dict` from each graph member instance and storing _that_, then re-creating the corresponding class instance on demand), the graph element classes in `pydot.core` were effectively un-subclassable. Even if you were to create subclasses of `pydot.core.Node` and `pydot.core.Graph` and add your `MyNode()` objects to your `MyGraph` instance, a call to `mygraph_instance.get_node()` would return `pydot.core.Node` objects instead of the `MyNode` objects you passed in.

To overcome this limitation and make the graph elements subclassable, add `_wrap_{node,edge,subgraph}` static methods to `pydot.core.Common`, and call `self._wrap_foo()` when constructing `Foo` objects to return, instead of directly calling the `Foo(obj_dict=)` constructor.

This means that a subclass `MyGraph` can be made to work with `MyNode` objects instead of `pydot.core.Node` objects simply by overriding `MyGraph._wrap_node()`, and so on for the other types/classes.

No API implications, this is an internal change that has no effect in `pydot.core.*` classes itself, it only makes the writing of subclasses easier.

Proposed / sketched out in https://github.com/pydot/pydot/pull/492#issuecomment-3115074502

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -565,6 +565,18 @@ class Common:
     def __setstate__(self, state: AttributeDict) -> None:
         self.obj_dict = state
 
+    @staticmethod
+    def _wrap_node(obj_dict: AttributeDict) -> Node:
+        return Node(obj_dict=obj_dict)
+
+    @staticmethod
+    def _wrap_edge(obj_dict: AttributeDict) -> Edge:
+        return Edge(obj_dict=obj_dict)
+
+    @staticmethod
+    def _wrap_subgraph(obj_dict: AttributeDict) -> Subgraph:
+        return Subgraph(obj_dict=obj_dict)
+
     def set_parent_graph(self, parent_graph: Common | None) -> None:
         self.obj_dict["parent_graph"] = parent_graph
 
@@ -904,7 +916,7 @@ class Edge(Common):
         indent_str = self.get_indent(indent, indent_level)
 
         if isinstance(src, FrozenDict):
-            sgraph = Subgraph(obj_dict=src)
+            sgraph = self._wrap_subgraph(obj_dict=src)
             edge = [
                 sgraph.to_string(
                     indent=indent, indent_level=indent_level, inline=True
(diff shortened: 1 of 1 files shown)


lkk7:
At first sight, it looks like a graphviz macOS-only bug 😮 It works on Graphviz 12. 

```
>>> dot --version
dot - graphviz version 13.1.0 (20250701.0955)

>>> dot -Tjpe test/graphs/url.dot 
[1]    31271 trace trap  dot -Tjpe test/graphs/url.dot
```

Then...

```
>> lldb -- dot -Tjpg test/graphs/url.dot
>> (lldb) run
Process 1651 launched: '/opt/homebrew/bin/dot' (arm64)
Process 1651 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BREAKPOINT (code=1, subcode=0x18425e034)
    frame #0: 0x000000018425e034 CoreFoundation`CFRelease.cold.1 + 16
CoreFoundation`CFRelease.cold.1:
->  0x18425e034 <+16>: brk    #0x1

CoreFoundation`__CFStringCollectionCopy.cold.1:
    0x18425e038 <+0>:  adrp   x8, 461
    0x18425e03c <+4>:  add    x8, x8, #0xea2 ; "*** __CFStringCollectionCopy() called with NULL ***"
    0x18425e040 <+8>:  adrp   x9, 451427
```

I don't know much, but `"*** __CFStringCollectionCopy() called with NULL ***"` looks bad. Then the Mac froze totally.

This is the minimal example that breaks it:
```
digraph G {
  name [URL=" "];
}
```

lkk7:
I had to close and re-open because I couldn't rerun checks (maybe the PR was too old).

Looks like the macOS bug is gone.

ferdnyc:
We should probably flip `macos` and `windows` back out of `can-fail: true` mode, now that they're reliable again. (This week.) That way, we won't miss the next time they get flaky.

(I noticed they were both still set that way when I was adding the freethreading jobs to the matrix.)

PR #497, opened 2025-09-03
Toward a thread-safe Pydot

### LINKED ISSUE #401: Data Race in parse_dot_data.

ferdnyc:
Recently, the Python community has been much abuzz about [free-threaded Python](https://docs.python.org/3/howto/free-threading-python.html), the experimental GIL-less builds of the interpreter available for Python 3.13 and 3.14. Running without the GIL has the promise of enabling greatly improved concurrency for Python programs, assuming they (and all of their package dependencies) are able to take advantage of it.

To help projects ready themselves for this new world, a `pytest-run-parallel` plugin is now available that will, instead of running the test suite distributed across multiple CPU cores (like `pytest-xdist` does), run the _same_ tests concurrently on multiple threads at once, poking at the package's thread-safety to see whether it trips up.

## What this PR is all about

Just to cut to the punchline here, I have a branch of Pydot which, after some major _organizational_ changes in `dot_parser` (but no functional changes), can fully pass the `pytest-run-parallel` gauntlet.

### Pydot thread-safety: myth or legend? (Spoiler: Neither!)

We've long claimed that Pydot isn't and can't be thread-safe, because the underlying Pyparsing itself doesn't make any claims to thread-safety. But, while it's true that Pyparsing **has** features that render it non-thread-safe (mostly, module-level configuration hooks), we don't actually **USE** any of those features. The lack of thread-safety in Pydot is completely down to the design of `dot_parser` itself, and it can be corrected.

### Pulling the threads

Those corrections won't be without cost, though. Making `dot_parser` thread-safe involves moving _all_ of the parse functions and parser bits into the `GraphParser` class (which I'd already created for other reasons), but making them **instance** variables and methods. Every run of `GraphParser.parser.parse_string()` now becomes `GraphParser().parser.parse_string()`, creating a new instance of the class for each and every parsing job. That's the only way it can possibly be thread-safe.

I know that _works_ (I've tested it), but I'm not entirely sure what the impact is performance-wise. That's what I want to test, now.

## The plan

So, for the moment this PR contains none of those changes, and only the changes required to **TEST** Pydot under the free-threaded `python3.13t` interpreter. The plan is to do this in stages:

1. Define a `py313t` tox environment that uses the free-threaded Python 3.13 to run our tests under `pytest-run-parallel`, instead of `pytest-xdist`.
2. Add an additional CI job (only on Ubuntu, for starters) that executes the freethreaded tests using `python3.13t`.
3. Let those tests run against the unmodified Pydot code. They will **FAIL MISERABLY**, as the package is not currently thread-safe. We know that. But, running the tests like this will give me a baseline for Pydot's _performance_ under current conditions.
4. Then, once we have that data, make the changes necessary for Pydot to **pass** the run-parallel tests in the freethreaded environment.
5. Compare the performance (based on simple test-suite runtime, for starters) before and after that change, to get a feel for what the impact will be of making Pydot thread-safe — in both the free-threaded environment, _**AND**_ the traditional GIL-enabled environment(s).

### Where we go from there

By taking the steps above, we can get a feel for what the cost will be of making Pydot thread-safe. If the cost seems reasonable, we can do it across-the-board. Otherwise, we may have to go a hybrid route, where thread-safety is only used when Pydot either detects it's running in a free-threaded environment, or the user configures it for thread-safety explicitly. We can't know until we have the data.

## TL;DR

In case it wasn't obvious (or the above was just too much information), this PR should be considered to have a big **DO NOT MERGE!!** sign on it, as it's only at Step 1 of a multi-phase process. CI runs of the intermediate steps are a necessity, so that we have sufficient data to analyze the remaining steps once they're taken.

### The phases

- [x] **PHASE 1:** Make changes to the CI environment to enable baseline testing of the unmodified Pydot code.
- [x] **PHASE 2:** Collect baseline CI results, for timing purposes
- [x] **PHASE 3:** Commit the package changes to the PR branch
- [ ] **PHASE 4:** Examine how the changes affect Pydot performance
- [ ] **PHASE 5:** Decide what changes the final PR should contain
- [ ] **PHASE 6:** Revise and merge (or abandon)
 
(Hopefully, eventually) Fixes #401


test/test_logging.py
```diff
@@ -12,6 +12,7 @@
 import pydot
 
 
+@pytest.mark.thread_unsafe(reason="Adjusts module-wide configs")
```
ferdnyc, line 15: (OK, a couple of thread-safety changes snuck in early — just to the tests, though, not to Pydot itself.)

- This test gets marked as non-thread-safe because it can never be. It diddles the Pydot module's global variables directly, just like the things we don't do to Pyparsing that would make _it_ unsafe.

test/conftest.py
```diff
@@ -44,6 +45,7 @@ def objdict() -> T.Dict[str, T.Any]:
         "edges": {},
         "subgraphs": {},
     }
+    return copy.deepcopy(obj)
```
ferdnyc, line 48: - And this fixture now makes a copy of the dictionary it supplies each time, because if it's passing the same dictionary object reference to each test, they'll stomp all over each other modifying it.

.github/workflows/CI.yml
```diff
@@ -44,6 +44,13 @@ jobs:
           - "3.11"
           - "3.10"
           - "3.9"
+        include:
+          - system: {os: ubuntu, can-fail: true}
+            python-version: "3.13t"
+
+    # Work arond tox-gh missing support for freethreaded variants
+    env:
+      TOX_GH_MAJOR_MINOR: ${{ matrix.python-version }}
```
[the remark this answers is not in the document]
ferdnyc: ..."arond"? 
```suggestion
    # Work around tox-gh missing support for freethreaded variants
    env:
      TOX_GH_MAJOR_MINOR: ${{ matrix.python-version }}
```

ferdnyc:
> Edit: I read it again. You meant that pyparsing is done already, because we don't use its non-thread-safe features at all. In that case, we just have to get rid of global variables and bad things in our own code.

Bingo, yes. Thanks for being able to pick the useful info out of my stream-of-consciousness rambling. 🙃

ferdnyc:
It's... "impressive"? how much slower the freethreaded interpreter is when you're _not_ making use of its features. Our non-thread-safe parser code runs almost _twice_ as slow, on Linux, as it does under regular Python 3.13. macOS and Windows are less affected, for some reason, but they still show severe slowdowns.

From what the docs say, Python 3.14 / 3.14t should improve on that significantly.

ferdnyc:
This Pr is gonna get hosed by the 100%-coverage PR anyway, I'll wait until that's merged and then fix up this one.

hugochinchilla:
I've been hit by this thread safety problem recently, nice to see work in progress to fix it.

IMHO I would not care if the changes make the parser slower, for my use-case I get a lot of benefits from being able to enable threading on my process that vastly surpass the downside of slower parsing, I know each one will have it's own use-cases and opinions on the matter.

Maybe a good middle ground is to provide two parser implementations on the library and let the user decide which one to use.

PR #500, opened 2025-09-10
100% coverage

ferdnyc:
This PR adds a raft of tests, to push our code coverage to 100%.

Admittedly, not every line is _executed_. I conservatively added `# pragma: no cover` and `# pragma: no branch` comments to the sections of code we're _not trying_ to cover, for a few reasons:

1. Some of it is test-failure handling code. It's **bad** if that's covered, it means tests are failing. (We measure coverage _of_ the tests, too, not just _with_ the tests.)
2. Some is defensive code, conditions or code paths that (hopefully) _can't_ ever be reached, but are in there "just in case". (Things like `None` checks on variables that aren't actually ever `None` in practice, etc...)
3. A couple of spots represent logic cul-de-sacs, dead code paths that the surrounding code steer away from ever reaching. We'll probably rip those out eventually.

The new tests here were written to specifically target the remaining actually-relevant code paths. Which (along with the pragmas) gets us to 100% coverage in CI. (Though not locally on my system, since a few code paths are Windows-only.)

Now we just have to keep it there!

### DIFF
--- a/test/test_api.py
+++ b/test/test_api.py
@@ -153,6 +153,16 @@ def test_get_subgraph() -> None:
     assert sout.get_name() == "sub"
 
 
+def test_cluster_casting() -> None:
+    c = pydot.Cluster("sg")
+    assert c.get_type() == "subgraph"
+    assert c.get_name() == "cluster_sg"
+    sg = pydot.Graph("sg")
+    sg_c = pydot.Cluster(obj_dict=sg.obj_dict)
+    assert sg_c.get_type() == "digraph"
+    assert sg_c.get_name() == "sg"
+
+
 def test_graph_defaults() -> None:
     g = pydot.Graph()
     g.set_graph_defaults(rank="TB", color="red")


ferdnyc:
@lkk7 (Huh, this turned out to be a good use for #500 after all!)

ferdnyc:
Oh, 45bc53c6ea1aac6d2804144aefde73efb3dd41b8 is an important one. It changes the return type of `Dot.create()` from `str` to `bytes`. With all of the knock-on effects that implies.

`create()` returns the binary results of generating `dot` output, which means 9 times out of 10 we're talking image data. It was _always_ `bytes`, the annotations  just lied. Now they're correct, and it even let me drop a couple of `type: ignore` comments because we're no longer passing the results of `create()` to a file handle returned from `open(foo, "wb")` for writing.

ferdnyc:
LOL. There's literally _one line_ of `core.py` missing from the coverage. Now I just have to find which one!

ferdnyc:
And, there it is. With a total of 226 lines of new tests added to `test_api.py`:

image

PR #540, opened 2026-04-15
WIP/PoC: Use functools.partialmethod for generated code, type-annotate with generated .pyi stubs

ferdnyc:
This is a proof-of-concept / experimental change to the way in which we generate all of the <code>set_<var>foo</var></code>, <code>get_<var>foo</var> </code>, <code>create_<var>foo</var></code> and <code>wrte_<var>foo</var></code> methods on the `pydot.core` graph-element classes.

##  Current state: shoehorned-in generated methods

Up until now, we've been creating initially unbound functions with an extra keyword argument that specified the format to use. That _worked_, as far as it goes, but not super well, and the extra argument was always there as part of the function signature. (In other words, right now with pydot you could do this, confusingly):

```python
import pydot
g = pydot.Graph("G")
g.set_bgcolor(5, _attr="penwidth")
```
...and the graph's `penwidth` would be set to `5`.

In addition, the generated functions can't have individual docstrings, and all of their `__code__` dunders point to the same location in the source: The `def` of the function they're all generated from:

```python
>>> import pydot
>>> pydot.core.Edge.set_penwidth
<function __generate_attribute_methods.<locals>.__setter at 0x7ff54e4c4930>
>>> pydot.core.Edge.set_penwidth.__code__
<code object __setter at 0x7ff54ec28430, file "/usr/lib/python3.14/site-packages/pydot/core.py", line 161>
>>> pydot.core.Graph.set_rank
<function __generate_attribute_methods.<locals>.__setter at 0x7ff54e4dd9b0>
>>> pydot.core.Graph.set_rank.__code__
<code object __setter at 0x7ff54ec28430, file "/usr/lib/python3.14/site-packages/pydot/core.py", line 161>
```

I've never _exactly_ been happy with that, but every other alternative seemed worse by comparison. Except, perhaps, this one.

## Partialmethod-generated method signatures

By using `functools.partialmethod` to generate the method signatures, the signatures for the generated methods are correct — <code>.get_<var>foo</var>()</code> takes no arguments, and <code>.set_<var>foo</var>()</code> takes only one.

### Wrappable `write` and `create` methods
For the <code>create_<var>foo</var></code> and <code>wrte_<var>foo</var></code> methods, this required providing an alternate signature for the base method that takes the `format` as its first positional parameter. That way, it can be supplied in the `partialmethod` call and removed from the signature. So, this PR also adds `_wrappable_create` and `_wrappable_write` methods to the `Dot` class with rearranged parameters. Those are then used when setting up all of the `partialmethod`-generated methods, and the resulting methods have the signature we want. (One where `format` is no part of the parameter list.)

## Still our circus, still our monkeys

To be clear, this isn't an ideal solution either — because the class is still being changed after-the-fact, we're still monkeypatching-in the extra methods. 

(A "correct" use of `functools.partialmethod` in a class definition would look like this):

```python
class Graph(Common)
    def set(self, name: str, value: Any) -> None:
        ...
    set_bgcolor = functools.partialmethod(set, "bgcolor")
```

We're not doing that, because we're (still) generating the method names instead of writing them all out in the source. So, it's still monkeypatching no matter how you slice it.

But the advantage of using `partialmethod` and having correct method signatures is that actual, correct signatures for these functions can now be _generated_ through inspection.

## `stubgen` to the... well, not "rescue", exactly...

After changing how the methods are being _generated_, I then ran the Mypy `stubgen` tool on the updated module definition for `pydot.core`, and what I got out was a promising starting point.

The generated `core.pyi` file produced by stubgen wasn't immediately correct. (Which is expected/normal. The docs are careful to say, "Stubgen generates _draft_ stubs. The auto-generated stub files often require some manual updates", and that was the case here. But the generated file was a _good_ start and (more importantly) contained 100% correct signatures for all of the generated methods right from the start. For example:

```python
class Graph(Common):
    # ...
    def get_levelsgap(self) -> Any: ...
    def set_levelsgap(self, value: Any) -> None: ...
    def get_label(self) -> Any: ...
    def set_label(self, value: Any) -> None: ...
    def get_mindist(self) -> Any: ...
    def set_mindist(self, value: Any) -> None: ...
    def get_bgcolor(self) -> Any: ...
    def set_bgcolor(self, value: Any) -> None: ...
    def get_clusterrank(self) -> Any: ...
    def set_clusterrank(self, value: Any) -> None: ...
    def get_margin(self) -> Any: ...
    def set_margin(self, value: Any) -> None: ...
    def get_concentrate(self) -> Any: ...
    def set_concentrate(self, value: Any) -> None: ...
    def get_mosek(self) -> Any: ...
    def set_mosek(self, value: Any) -> None: ...
    # ...
```
(The order seems to be totally random, and changes from run to run. That's annoying and I would love to find a way to fix it.)

This means that, if the `.pyi` file is included with the Pydot distribution, editors that use type annotations for completion, validation, or inline documentation will be able to see, display, and use the signatures for all of those methods, something that previously didn't work particularly well.

## Other changes

As I was working on this, I discovered (when it completely broke my generated stubs) that the `OUTPUT_FORMATS` set contained an element `"plain-ext"`, and we've been dutifully generating methods on the `Dot` class named `write_plain-ext` and `create_plain-ext`. I don't think I have to explain why that's a problem. So, there's a commit here that just deletes `plain-ext` from the set.


ferdnyc:
## Where does this go from here?

Because the _methods_ are generated from lists (well, sets) included in the source, I'd really feel better if the _stubs_ were also generated, so that they're always in sync with the code and we don't have to worry about a change being made in `core.py` that's not reflected in the corresponding `.pyi` file.

The `stubtest` tool (also part of Mypy) can help watch for that kind of thing by checking the stubs against the code, but even if it finds a problem it will have to be **corrected** manually, because re-generating the stubs with `stubgen` will lose the manual corrections I made so they're actually usable.

(Which, to be clear, are _not_ extensive. They're mostly small things, like):

* Updating the type annotation for all of the static `set`s from `: set` to `: set[str]` — though we may be able to automate that part by adding the same annotations to `core.py`.
* Adding in the backwards-compatible type aliases like `EdgeDefinition = Union[EdgeEndpoint, Node, Subgraph, Cluster]` ­— when `stubgen` can't find definitions it knows how to process, it just assumes `EdgeEndpoint` and `EdgeDefinition` can be imported from `typing`. (Which is a bit weird, as a default/fallback, but that's how it works.)
* Adding a `# type: ignore` comment to the `FrozenDict.__hash__` definition in the `classes.pyi` stub. Otherwise, the type-checking tools will complain (correctly) that our working `__hash__` method is incompatible with the `dict` base class's definition, since `dict`s are not hashable. (That's one of the reasons I don't love  using `dict` as the base class for `FrozenDict`, but that's an entirely separate issue.)

While it would be preferable in some ways if the `.pyi` files weren't checked in to the repo at all, and we just generated them during the packaging process, the manual edits needed make that impossible. And anyway, that wouldn't help anyone who's using a checked-out `pydot` from source. So, we do want the `.pyi` files to be _there_. But we need some level of automation to validate them (and ideally, make updates as well) to ensure they don't fall out of sync.

ferdnyc:
> * Updating the type annotation for all of the static `set`s from `: set` to `: set[str]` — though we may be able to automate that part by adding the same annotations to `core.py`.

Damn. I just tested, and that (surprisingly) doesn't work. Even if `core.py` contains this:

```python
CLUSTER_ATTRIBUTES: set[str] = {
    "K", "URL", "bgcolor", "color", "colorscheme",
    "fillcolor", "fontcolor", "fontname", "fontsize", "label", "labeljust",
    "labelloc", "lheight", "lp", "lwidth", "nojustify", "pencolor",
    "penwidth", "peripheries", "sortv", "style", "target", "tooltip"
}
```
the `stubgen`-erated annotation is:
```python
CLUSTER_ATTRIBUTES: set
```
But now, there's an extra module-level annotation in the `.pyi`:
```python
__conditional_annotations__: set
```

...I have _no idea_ what that's all about.

ferdnyc:
One open question is whether we could leverage the `.pyi` files to more precisely _type_ our convenience functions, as well. It'd be great if there was some way to arrange things so that `stubgen` produces signatures like this:

```python
class Graph(Common):
    def get_levelsgap(self) -> int | float: ...
    def set_levelsgap(self, value: int | float) -> None: ...
    def get_label(self) -> str: ...
    def set_label(self, value: str) -> None: ...
    def get_mindist(self) -> int | float: ...
    def set_mindist(self, value: int | float) -> None: ...
    def get_bgcolor(self) -> str: ...
    def set_bgcolor(self, value: str) -> None: ...
    def get_clusterrank(self) -> Literal[local] | Literal[global] | Literal[none]: ...
    def set_clusterrank(
        self,
        value: Literal[local] | Literal[global] | Literal[none]
    ) -> None: ...
    def get_margin(self) -> int | float | str: ...
    def set_margin(self, value: int | float | str) -> None: ...
    def get_concentrate(self) -> bool: ...
    def set_concentrate(self, value: bool) -> None: ...
```
...but that might be a bit overambitious. And could even cause other problems with type-checking, since `Graph.set` will still take a `value` type of `Any`, and `Graph.get` will still return `Any`.

(**Edit:** Oh, right — and that wouldn't exactly work anyway, without type-validation improvements to the rest of the code. Because, if our parser encounters a graph definition like this:

```dot
graph G {
  concentrate=true
}
```

It's going to store the _string_ value `"true"` in the graph's `obj_dict["attributes"]["clusterrank"]`, NOT the boolean value `True`.)

ferdnyc:
Hmm. I just had a thought. Well, a series of thoughts.

The reason the stubs have to be so complex is, if there's a `.pyi` file for a module, the type annotations actually _in_ the module will basically be ignored, so ALL of the type information has to be in the stubs.

Which got me thinking, maybe it would be better to move the _generated_ code to a separate module, defined as mixin classes for the `core.py` classes. That way, typing for `mixins.py` could be generated by `stubgen` as `mixins.pyi`, but there wouldn't have to be a `core.pyi` and (IIUC) its own internal type information would still be used.

But then I realized, if we're going to generate a `mixins.pyi` stub file from a `mixins.py` that _generates_ class methods, why not just **generate** `mixins.py` ITSELF? It's not much more of a step up from what we're already doing, to write a script that generates a Python source file like this:

```python
# (SPDX headers)

from __future__ import annotations

from typing import Any


class NodeMixin:
    # To get the class to typecheck on its own. I'm not 100% sure if these
    # are necessary, I think mypy may have special handling for mixin classes.
    def get(self, key: str) -> Any:
        raise NotImplementedError

    def set(self, key: str, value: Any) -> None:
        raise NotImplementedError

    # Then, the real convenience methods
    def get_penwidth(self) -> Any:
        """Get the ``penwidth`` attribute for the ``Node``"""
        return self.get("penwidth")

    def set_penwidth(self, value: Any) -> None:
        """Set the ``penwidth`` attribute for the ``Node``"""
        return self.set("penwidth", value)

    def get_pos(self) -> Any:
        """Get the ``pos`` attribute for the ``Node``"""
        return self.get("pos")

    def set_pos(self, value: Any) -> None:
        """Set the ``pos`` attribute for the ``Node``"""
        return self.get("pos", value)
````

...You get the idea. This way, we don't need any stubs at all, because all of the methods would _**actually have**_ statically-analyzable, type-annotated, explicit lines of code for the introspection APIs to point to, and for mypy and etc. to parse. They'd even have proper (if trivial) docstrings, unlike the current generated methods.

The definitions in `core.py` would just change to:

```python
# ...
from pydot.mixins import (
    ClusterMixin, DotMixin, EdgeMixin, GraphMixin, NodeMixin
)
# ...
class Node(NodeMixin, Common):  # Or the other way around?
    # ...                       # I can never remember.


class Edge(EdgeMixin, Common):
    # ...

# (and so on)
```

The script to generate `mixins.py` could live in the repo, _not_ in the `src/pydot/` dir (because it's a utility script, not part of the package sources), and the CI could run it and compare it to the checked-in copy to ensure that nobody forgets to regenerate it whenever there are changes.

I'm gonna give that a try.

ferdnyc:
> ```python
> class Node(NodeMixin, Common):  # Or the other way around?
>     # ...                       # I can never remember.
> ```

...It was, indeed, the other way around.

I'm going to close this PR and open a new one implementing the new approach, which I've migrated to a new branch name. I want to preserve the code for this (aborted) attempt for posterity, and as a cautionary tale.

PR #541, opened 2026-04-18
WIP: Generate mixin classes with explicit code for convenience methods

ferdnyc:
This PR takes a revamped approach to handling the generated convenience methods for the pydot classes, evolved from my first attempt in the previous (now-closed) PR #540. (Effectively, throwing away everything I was doing there in favor of a completely different approach that is — I hope — better.)

In _this_ PR, a new script `generate_mixins.py` is added to the root of the repository, and used to generate the file `src/pydot/mixins.py`. `pydot.mixins` contains (for the moment) mixin classes that implement the attribute `get_` and `set_` methods for the `pydot.core` classes as explicit code, with source file locations and docstrings and the works. Effectively, instead of generating _methods_, I'm now generating _source code_ for those methods, which is more conducive to static analysis and type-checking.

The mixin classes (`CommonMixin` —  I did some deduplicating of the attribute lists and made shared methods inheritable, `NodeMixin`, `EdgeMixin`, `GraphMixin`, and `ClusterMixin`) are used as parent classes to the corresponding `pydot.core.` graph-element classes, and they provide all of the <code>get_<var>foo</var></code> and <code>set_<var>foo</var></code> methods that we used to be generating at runtime.

The `mypy-check` tox job still passes, and all of the tests still pass, but now we have things like this:

```python
>>> import pydot
>>> help(pydot.Graph.set_bgcolor)
Help on function set_bgcolor in module pydot.mixins:

set_bgcolor(self, value: Any) -> None
    Set the ``bgcolor`` attribute for the graph element.

>>> help(pydot.Node)
Help on class Node in module pydot.core:

class Node(Common, pydot.mixins.NodeMixin)
 |  Node(name: str = '', obj_dict: AttributeDict | None = None, **attrs: Any) -> None
 |
 |  A graph node.
 |
 |  This class represents a graph's node with all its attributes.
 |
 |  node(name, attribute=value, ...)
 |
 |  name: node's name
 |
 |  All the attributes defined in the Graphviz dot language should
 |  be supported.
 |
 |  Method resolution order:
 |      Node
 |      Common
 |      pydot.mixins.CommonMixin
 |      pydot.mixins.NodeMixin
 |      builtins.object
 |
 |  Methods defined here:
 |
 |  __init__(
 |      self,
 |      name: str = '',
 |      obj_dict: AttributeDict | None = None,
 |      **attrs: Any
 |  ) -> None
 |      Initialize self.  See help(type(self)) for accurate signature.
 |
 |  __str__(self) -> str
 |      Return str(self).
 |
 |  add_style(self, style: str) -> None
 |
 |  get_name(self) -> str
 |      Get the node's name.
 |
 |  get_port(self) -> str | None
 |      Get the node's port.
 |
 |  set_name(self, node_name: str | None) -> None
 |      Set the node's name.
 |
 |  to_string(self, indent: Any = '', indent_level: int = 1) -> str
 |      Return string representation of node in DOT language.
 |
 |  ----------------------------------------------------------------------
 |  Methods inherited from Common:
 |
 |  __getstate__(self) -> AttributeDict
 |      Helper for pickle.
 |
 |  __setstate__(self, state: AttributeDict) -> None
 |
 |  attrs_string(self, prefix: str = '') -> str
 |      Format the current attributes list for output.
 |
 |      The `prefix` string will be prepended if and only if some
 |      output is generated.
 |
 |  formatted_attr_list(self) -> list[str]
 |      Return a list of the class's attributes as formatted strings.
 |
 |  get(self, name: str) -> Any
 |      Get an attribute value by name.
 |
 |      Given an attribute 'name' it will get its value.
 |      There's always the possibility of using the methods:
 |
 |          get_'name'()
 |
 |      which are defined for standard graphviz attributes.
 |
 |  get_attributes(self) -> AttributeDict
 |      Get attributes of the object
 |
 |  get_parent_graph(self) -> Graph | None
 |
 |  get_sequence(self) -> int | None
 |      Get sequence
 |
 |  get_top_graph_type(self, default: str = 'graph') -> str
 |      Find the topmost parent graph type for the current object.
 |
 |  set(self, name: str, value: Any) -> None
 |      Set an attribute value by name.
 |
 |      Given an attribute 'name' it will set its value to 'value'.
 |      There's always the possibility of using the methods:
 |
 |          set_'name'(value)
 |
 |      which are defined for standard graphviz attributes.
 |
 |  set_parent_graph(self, parent_graph: Common | None) -> None
 |
 |  set_sequence(self, seq: int) -> None
 |      Set sequence
 |
 |  ----------------------------------------------------------------------
 |  Static methods inherited from Common:
 |
 |  get_indent(indent: Any, indent_level: int) -> str
 |
 |  ----------------------------------------------------------------------
 |  Data and other attributes inherited from Common:
 |
 |  __annotations__ = {}
 |
 |  obj_dict = {}
 |
 |  ----------------------------------------------------------------------
 |  Methods inherited from pydot.mixins.CommonMixin:
 |
 |  get_URL(self) -> Any
 |      Get the ``URL`` attribute for the graph element.
 |
 |  get_colorscheme(self) -> Any
 |      Get the ``colorscheme`` attribute for the graph element.
 |
 |  get_fontcolor(self) -> Any
 |      Get the ``fontcolor`` attribute for the graph element.
 |
 |  get_fontname(self) -> Any
 |      Get the ``fontname`` attribute for the graph element.
 |
 |  get_fontsize(self) -> Any
 |      Get the ``fontsize`` attribute for the graph element.
 |
 |  get_label(self) -> Any
 |      Get the ``label`` attribute for the graph element.
 |
 |  get_nojustify(self) -> Any
 |      Get the ``nojustify`` attribute for the graph element.
 |
 |  get_target(self) -> Any
 |      Get the ``target`` attribute for the graph element.
 |
 |  set_URL(self, value: Any) -> None
 |      Set the ``URL`` attribute for the graph element.
 |
 |  set_colorscheme(self, value: Any) -> None
 |      Set the ``colorscheme`` attribute for the graph element.
 |
 |  set_fontcolor(self, value: Any) -> None
 |      Set the ``fontcolor`` attribute for the graph element.
 |
 |  set_fontname(self, value: Any) -> None
 |      Set the ``fontname`` attribute for the graph element.
```
...and so on.

### Left to do

- [ ] Write the generator for a `DotMixin` class containing the `write_` and `create_` methods for the output formats.
- [ ] Add a CI check that runs `generate_mixins.py` and compares the output to the checked-in `src/pydot/mixins.py`. Any deviation should fail the CI run, since somebody forgot to regenerate and check in an updated `mixins.py` after making changes.
- [ ] Also generate a `test/test_mixins.py` with tests for all of the convenience methods, to restore our coverage numbers.

### Other changes

* The <code><var>FOO</code>_ATTRIBUTES</code> sets are migrated from `core.py` into a new `constants.py` file, because I need to be able to import them into `generate_mixins.py` _without_ importing any of the rest of `pydot`.

    (If I'm potentially going to be overwriting one of the source files, having it loaded isn't a great idea. And importing any part of `pydot` will import `pydot.mixins`, because of the way we do symbol hoisting into the `__init__.py` module. So, `generate_mixins.py` adds `src/pydot` to `sys.path`, and imports the symbols directly `from constants`, which avoids triggering `src/pydot/__init__.py`.)

* As I was working on the approach in #540, I discovered (when it completely broke my generated type stubs) that the `OUTPUT_FORMATS` set contained an element `"plain-ext"`, and we've been dutifully generating methods on the `Dot` class named `write_plain-ext` and `create_plain-ext`. I don't think I have to explain why that's a problem. So, there's a commit here that just deletes `plain-ext` from the set, since it would similarly foul a generated `mixins.py` that contained unparseable code like this:

    ```python
    class DotMixin:
        def write_plain-ext(...):
    ```


ferdnyc:
CI failure was a weird Chocolatey hiccup, seems fine now.

ferdnyc:
AUGH, my coverage! :cry: 

Now that there's actually explicit source for all of these methods, I guess I have to either also generate _tests_ for all of them, or exempt the `mixins.py` file from coverage checking.

I'll probably go the "generate a tests file" route, because I'd hate to be claiming 100% coverage when we're ignoring the entire contents of a 1500-line file.

ferdnyc:
Oh, yes, I made the `obj_dict` member of `Common` explicit (it's part of the class definition, with an explicit type annotation and everything), because we really should've been doing that always, and when I was pursuing the type-stubs approach, not having it there meant it wasn't part of the stub file. It's not strictly _necessary_ in the current approach, but it's still a good idea.

PR #542, opened 2026-04-21
100% type coverage and improved annotations

jorenham:
I noticed on the typestats dashboard ([url](https://jorenham.github.io/typestats/dashboard/report/#pydot)) that `pydot` has 86% type coverage. So I figured I might as well help out and bring that to 100% :)

Before:

```bash
$ uvx typestats check pydot
untyped (44):
pydot/core.py:279        re_numeric
pydot/core.py:280        re_dbl_quoted
pydot/core.py:281        re_html
pydot/core.py:283        id_re_alpha_nums
pydot/core.py:284        id_re_alpha_nums_with_ports
pydot/core.py:287        id_re_with_port
pydot/core.py:1658       Dot.formats
pydot/core.py:1659       Dot.prog
pydot/dot_parser.py:56   P_AttrList.attrs
pydot/dot_parser.py:77   DefaultStatement.default_type
pydot/dot_parser.py:78   DefaultStatement.attrs
pydot/dot_parser.py:264  GraphParser.colon
pydot/dot_parser.py:265  GraphParser.lbrace
pydot/dot_parser.py:266  GraphParser.rbrace
pydot/dot_parser.py:267  GraphParser.lbrack
pydot/dot_parser.py:268  GraphParser.rbrack
pydot/dot_parser.py:269  GraphParser.equals
pydot/dot_parser.py:270  GraphParser.comma
pydot/dot_parser.py:271  GraphParser.semi
pydot/dot_parser.py:272  GraphParser.minus
pydot/dot_parser.py:275  GraphParser.strict_
pydot/dot_parser.py:276  GraphParser.graph_
pydot/dot_parser.py:277  GraphParser.digraph_
pydot/dot_parser.py:278  GraphParser.subgraph_
pydot/dot_parser.py:279  GraphParser.node_
pydot/dot_parser.py:280  GraphParser.edge_
pydot/dot_parser.py:283  GraphParser.identifier
pydot/dot_parser.py:285  GraphParser.double_quoted
pydot/dot_parser.py:291  GraphParser.concat_string
pydot/dot_parser.py:295  GraphParser.ID
pydot/dot_parser.py:302  GraphParser.float_number
pydot/dot_parser.py:306  GraphParser.node_id
pydot/dot_parser.py:310  GraphParser.a_list
pydot/dot_parser.py:313  GraphParser.attr_list
pydot/dot_parser.py:316  GraphParser.node_stmt
pydot/dot_parser.py:323  GraphParser.default_stmt
pydot/dot_parser.py:325  GraphParser.stmt_list
pydot/dot_parser.py:326  GraphParser.graph_stmt
pydot/dot_parser.py:333  GraphParser.subgraph
pydot/dot_parser.py:337  GraphParser.edgeop
pydot/dot_parser.py:339  GraphParser.edge_stmt
pydot/dot_parser.py:356  GraphParser.parser
pydot/dot_parser.py:365  GraphParser.single_line_comment
pydot/exceptions.py:25   Error.value

coverage:   86.03%
typable:    315
typed:      226
any:         45
```

After:

```bash
$ uvx typestats check pydot
coverage:   100.00%
typable:    327
typed:      283
any:         44
```

---

While I was working on this, I was also using [pyrefly](https://github.com/facebook/pyrefly), which helped spot some other minor opportunities for improvements here and there. Some of those I included in here. Others, such as making `FrozenDict` a [generic type](https://typing.python.org/en/latest/reference/generics.html) will have to wait #539 to avoid it from becoming a mess.

Either way, if you're open to it, maybe it's an idea to (also) use pyrefly as type-checker? Personally I prefer it over mypy, and in my experience pyrefly and mypy can play very well together, complementing each other even. So if you're up for it I could add pyrefly as a follow-up, or perhaps even in this PR?

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -952,6 +952,7 @@ class Edge(Common):
 
 __generate_attribute_methods(Edge, EDGE_ATTRIBUTES)
 
+
 class Graph(Common):
     """Class representing a graph in Graphviz's dot language.
 
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -15,12 +15,12 @@ import re
 import subprocess
 import sys
 import warnings
-from typing import TYPE_CHECKING, Any, Final, Sequence, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Any, Final, Sequence, Union, cast
 
 if TYPE_CHECKING:
     # `typing_extensions` is always available in `TYPE_CHECKING` blocks,
     # even if not  installed
-    from typing_extensions import TypeAlias
+    from typing_extensions import Self, TypeAlias
 
 import pydot
 from pydot._vendor import tempfile
@@ -952,9 +952,6 @@ class Edge(Common):
 
 __generate_attribute_methods(Edge, EDGE_ATTRIBUTES)
 
-TGraph = TypeVar("TGraph", bound="Graph")
-
-
 class Graph(Common):
     """Class representing a graph in Graphviz's dot language.
 
@@ -1134,7 +1131,7 @@ class Graph(Common):
         self.obj_dict["current_child_sequence"] = seq + 1
         return seq
 
-    def __enter__(self: TGraph) -> TGraph:
+    def __enter__(self) -> Self:
         """Enter the runtime context for this graph.
 
         Enables a nested construction style:
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -15,7 +15,12 @@ import re
 import subprocess
 import sys
 import warnings
-from typing import Any, Sequence, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Any, Final, Sequence, TypeVar, Union, cast
+
+if TYPE_CHECKING:
+    # `typing_extensions` is always available in `TYPE_CHECKING` blocks,
+    # even if not  installed
+    from typing_extensions import TypeAlias
 
 import pydot
 from pydot._vendor import tempfile
@@ -25,7 +30,7 @@ _logger = logging.getLogger(__name__)
 _logger.debug("pydot core module initializing")
 
 # fmt: off
-GRAPH_ATTRIBUTES = {
+GRAPH_ATTRIBUTES: Final = {
     "Damping", "K", "URL", "aspect", "bb", "bgcolor",
     "center", "charset", "clusterrank", "colorscheme", "comment", "compound",
     "concentrate", "defaultdist", "dim", "dimen", "diredgeconstraints",
@@ -45,7 +50,7 @@ GRAPH_ATTRIBUTES = {
 }
 
 
-EDGE_ATTRIBUTES = {
+EDGE_ATTRIBUTES: Final = {
     "URL", "arrowhead", "arrowsize", "arrowtail",
     "color", "colorscheme", "comment", "constraint", "decorate", "dir",
     "edgeURL", "edgehref", "edgetarget", "edgetooltip", "fontcolor",
@@ -61,7 +66,7 @@ EDGE_ATTRIBUTES = {
 }
 
 
-NODE_ATTRIBUTES = {
+NODE_ATTRIBUTES: Final = {
     "URL", "color", "colorscheme", "comment",
     "distortion", "fillcolor", "fixedsize", "fontcolor", "fontname",
     "fontsize", "group", "height", "id", "image", "imagescale", "label",
@@ -74,7 +79,7 @@ NODE_ATTRIBUTES = {
 }
 
 
-CLUSTER_ATTRIBUTES = {
+CLUSTER_ATTRIBUTES: Final = {
     "K", "URL", "bgcolor", "color", "colorscheme",
     "fillcolor", "fontcolor", "fontname", "fontsize", "label", "labeljust",
     "labelloc", "lheight", "lp", "lwidth", "nojustify", "pencolor",
@@ -83,7 +88,7 @@ CLUSTER_ATTRIBUTES = {
 # fmt: on
 
 
-OUTPUT_FORMATS = {
+OUTPUT_FORMATS: Final = {
     "canon",
     "cmap",
     "cmapx",
@@ -123,7 +128,7 @@ OUTPUT_FORMATS = {
 }
 
 
-DEFAULT_PROGRAMS = {
+DEFAULT_PROGRAMS: Final = {
     "dot",
     "twopi",
     "neato",
@@ -136,7 +141,7 @@ DEFAULT_PROGRAMS = {
 class frozendict(FrozenDict):
     """Deprecated alias for pydot.classes.FrozenDict."""
 
-    def __init__(self, *args: Any, **kwargs: Any):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
         warnings.warn(
             f"{self.__class__.__name__} is deprecated. "
             "Use pydot.classes.FrozenDict instead.",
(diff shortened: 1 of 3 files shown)


src/pydot/core.py
```diff
@@ -25,7 +30,7 @@
 _logger.debug("pydot core module initializing")
 
 # fmt: off
-GRAPH_ATTRIBUTES = {
+GRAPH_ATTRIBUTES: Final = {
```
ferdnyc, line 33: Huh. I don't really get `Final` as a type for module-level variables, to be honest. I get it in classes (you'll get an error if you try to modify the property in a subclass), but the documentation note that "There is no runtime checking of these properties" makes me nervous typing a _collection_ as `Final`. 

I'm not dead-set against it, merely on the fence. I could still be sold on it.

ferdnyc: Although, from experimentation, it appears the "no runtime checking" note refers _only_ to the `Final` aspect, so that's less concerning:

```python
>>> from typing import Final, reveal_type
>>> DEFAULT_PROGRAMS = {'dot', 'neato'}
>>> reveal_type(DEFAULT_PROGRAMS)
Runtime type is 'set'
{'dot', 'neato'}
>>> DEFAULT_PROGRAMS: Final = {'dot', 'neato'}
>>> reveal_type(DEFAULT_PROGRAMS)
Runtime type is 'set'
{'dot', 'neato'}
>>> DEFAULT_PROGRAMS: Final[set[str]] = {'dot', 'neato'}
>>> reveal_type(DEFAULT_PROGRAMS)
Runtime type is 'set'
{'dot', 'neato'}
```

src/pydot/core.py
```diff
@@ -1651,10 +1660,14 @@ class Dot(Graph):
...
 
+    shape_files: list[str]
+    formats: set[str]
+    prog: str
+
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
 
-        self.shape_files: list[str] = []
+        self.shape_files = []
```
ferdnyc, line 1692: I'm curious why you'd want to _remove_ that annotation. The idea is that, when elements are later being added to the list, if anyone ever writes code that tries something like `self.shape_files.append(12)` they'll get a type error. But that requires the explicit contents-type annotation, no?

src/pydot/core.py
```diff
@@ -1873,4 +1886,4 @@ def create(
 
 
 # Type alias for forward-referenced type
-EdgeDefinition = Union[EdgeEndpoint, Node, Subgraph, Cluster]
+EdgeDefinition: TypeAlias = Union[EdgeEndpoint, Node, Subgraph, Cluster]
```
ferdnyc, line 1911: When we tried that before, it broke things in older Python versions, but I think that was before we added `from __future__ import annotations` to the modules. So maybe it's OK now.

src/pydot/dot_parser.py
```diff
@@ -261,88 +265,94 @@ class GraphParser:
...
-    minus = Literal("-")
+    colon: ClassVar[Literal] = Literal(":")
+    lbrace: ClassVar[Literal] = Literal("{")
+    rbrace: ClassVar[Literal] = Literal("}")
+    lbrack: ClassVar[Literal] = Literal("[")
+    rbrack: ClassVar[Literal] = Literal("]")
+    equals: ClassVar[Literal] = Literal("=")
+    comma: ClassVar[Literal] = Literal(",")
+    semi: ClassVar[Literal] = Literal(";")
+    minus: ClassVar[Literal] = Literal("-")
```
ferdnyc, line 276: Honestly, rather than adding type annotations I would prefer to just get _rid_ of these entirely. Unlike the `CaselessKeyword` strings, they're wholly unnecessary aliases which can be substituted with just the string itself. The variables only obfuscate the parsing definitions, IMHO.

You may notice there are parts of the code that use explicit punctuation literals instead of the variables, mostly because it's newer and I wrote it.

Even when suppressing the punctuation, it's more readable (again, IMHO) to write:
```python
attr_list = OneOrMore(
    Suppress("[") + Optional(a_list) + Suppress("]")
)

graph_stmt = Group(
     Suppress("{")
     + Optional(stmt_list)
     + Suppress("}")
     + Optional(";").suppress()
)
```

Rather than the way we have it now:
```python
attr_list = OneOrMore(
    lbrack.suppress() + Optional(a_list) + rbrack.suppress()
)

graph_stmt = Group(
    lbrace.suppress()
    + Optional(stmt_list)
    + rbrace.suppress()
    + Optional(semi.suppress())
)
```

And when it's _not_ being suppressed, I vastly prefer avoiding this:

```python
a_list = OneOrMore(
    ID + Optional(equals + righthand_id) + Optional(comma.suppress())
)

assignment = ID + equals + righthand_id
```

in favor of the 100% equivalent:

```python
a_list = OneOrMore(
    ID + Optional("=" + righthand_id) + Optional(",").suppress()
)

assignment = ID + "=" + righthand_id
```

src/pydot/dot_parser.py
```diff
@@ -261,88 +265,94 @@ class GraphParser:
...
         double_quoted, delim="+", min=2, combine=False
     )
 
-    ID = (
+    ID: ClassVar[ParserElement] = (
         concat_string("concat")
         | double_quoted
         | identifier("ident")
         | HTML().set_results_name("html")
     ).set_parse_action(push_ID)
```
ferdnyc, line 306: I suppose, for the moment, defining these as `ClassVar`s is valid enough. Although if we ever do pull the trigger on #497, part of making the parser thread-safe would be to move the parser creation into the class's `__init__` so that all of the `ParserElement`s _are_ instance variables. That way, each parser instance is isolated from any others running on a different thread. So, longer-term the goal may end up being to eliminate any `ClassVar`s.

But, like I said, for the moment it's not technically wrong.

src/pydot/exceptions.py
```diff
@@ -21,6 +23,8 @@ class PydotException(Exception):
 class Error(PydotException):
     """General error handling class."""
 
+    value: Final[str]
+
```
ferdnyc: That feels wrong to me. Who's to say that a subclass won't want to override `value`? They shouldn't be prevented from doing this:

```python
class TaggedError(Error):
    def __init__(self, value: str, tag: str) -> None:
        super().__init__(value)
        self.value += f"[{tag}]"
```
Yes, _technically_ they could implement it like this, to avoid violating the `Final` constraint:
```python
class TaggedError(Error):
    def __init__(self, value: str, tag: str) -> None:
        super().__init__(f"{value}[{tag}]")
```
...but I don't see why we'd make them jump through hoops like that unnecessarily.

(Anyway this is just a contrived example, obviously. There may be other real-world scenarios where the `Final` isn't as easy to work around.)

ferdnyc (commented):
Thanks @jorenham, no question our type annotations can definitely be improved.

I've never looked at pyrefly myself, so for now I'll withhold any opinion on adding it as an additional type checker. Certainly, adding a tox environment to run it is simple enough, even if it's an optional one to start.

src/pydot/core.py
jorenham: > but the documentation note that "There is no runtime checking of these properties" 

That's right; type annotations aren't enforced at runtime. It's purely meant for static type-checkers.

---

> makes me nervous typing a _collection_ as `Final`.

`Final` is a rather confusing qualifier IMHO. Because you might think that it means "cannot be changed" (like a `const` in languages in rust), but it's actually [specified](https://typing.python.org/en/latest/spec/qualifiers.html#uppercase-final) as follows:

> The `typing.Final` [type qualifier](https://typing.python.org/en/latest/spec/glossary.html#term-type-qualifier) is used to indicate that a variable or attribute should not be reassigned, redefined, or overridden.

To it can still be mutated, but it cannot be re-assigned.

src/pydot/core.py
jorenham: I didn't **re**move this annotations, I *moved* to line 1663 :)

src/pydot/core.py
jorenham: Yea that makes sense: `typing.TypeAlias` was introduced in Python 3.10. By importing it in a `TYPE_CHECKING` block (line 23) in combination with the `from __future__ import annotations`, we avoid this from raising at runtime.

src/pydot/dot_parser.py
jorenham: Yea that makes sense. TBH it feels a bit out-of-scope here, so maybe it would be cleaner to address this separately, and leave this for now?

src/pydot/exceptions.py
jorenham: Hmm yea that seems like a valid use-case. I'll get rid of this `Final` :ok_hand:

jorenham:
Thanks for the quick review @ferdnyc; much appreciated.

lkk7:
@jorenham Thanks. Sorry it took so long.

jorenham:
Thanks for reviewing!

PR #546, opened 2026-05-09
[Experimental] Add `uv` package manager support

lkk7:
- Add `uv` support
- Use it in CI
- Add Python 3.15 testing

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -43,6 +43,7 @@ jobs:
           - {os: windows, can-fail: false}
           - {os: macos, can-fail: false}
         python-version:
+          - "3.15"
           - "3.14"
           - "3.13"
           - "3.12"
@@ -54,11 +55,16 @@ jobs:
       - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           fetch-depth: 0
-      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+      - name: Install uv and Set Python Version
+        uses: astral-sh/setup-uv@v5
         with:
+          enable-cache: true
+          cache-dependency-path: "pyproject.toml"
           python-version: "${{ matrix.python-version }}"
-          allow-prereleases: true
-          cache: "pip"
+      - name: "Verify Python Environment"
+        run: |
+          echo "Python used by uv:"
+          uv run python --version
       - name: "Set up environment"
         shell: bash
         run: |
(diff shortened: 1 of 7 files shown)


lkk7:
`uv` is a package manager which makes some things easier, like testing with a specific Python version, running some one-time-use tools and automatically cleaning up, and so on. It's way faster than others, too (it might be imperceptible in our CI because most of the time is taken by tests)

@ferdnyc thoughts? Do you know it? Like it / hate it? That's just an idea.

Edit after some thinking:

TBH, there's three parts of uv:

- As a user: I can just use uv to install pydot. Or I can install it traditionally. 
- As a code tinkerer: I can run tests with a chosen version quickly with uv. Testing some new scanner or linter or tool on the repo, that's very quick too. I can also do it traditionally.
- As a contributor who changes dependencies: I HAVE to update uv.lock or else I make the repo state inconsistent.

So I'm not sure if it's worth it for such a low-dependency (pyparsing only) project like pydot. Because the first two elements of uv are easy to use if someone wants them, whether we support it or not. And the third one locks every future dev into uv. While the benefit is... Possibly not that great with one dependency. Well, there are also  dev dependencies like the formatter/linter/typechecker. 
So maybe it's best not to force any lock files into this repository, but consider the usage in CI.

ferdnyc:
@lkk7 I "know of" uv more than _know_ it; [huggingface/transformers](https://github.com/huggingface/transformers/) is the most recent project I've looked at that uses it. I've heard good things about it, though I confess the "rewrite all the things in Rust!" trend has reached a critical mass where it gives me pause. Because:

1. There is real value in dogfooding your own implementation language, especially when it comes to build and install tools. I worry about a world where Python codebases are maintained using entirely non-Python tools.
2. The arguments for Rustination, beyond the standard "Rust is just better" stuff, often invoke speed as a major benefit. Linting tools like Black and etc. can be extremely slow, and unquestionably benefit from acceleration. (Ruff is admittedly warp-speed faster, compared to the Python tools it replaces.) But I've never felt either pip or tox were _slow_ or impeding my repo tasks in a way that they **need** _speeding up_ with uv.
3. I'm wary of the gatekeeping that can arise from a lot of Python projects becoming beholden to some Rust tools, especially ones written by a corporate team, and **especially** if it's the _SAME_ corporate team. uv is another project from Astral, the developers of Ruff.<sup>1</sup>

But, you know... all that being said, we already use Ruff here, so there really isn't much philosophical ground to be lost in adopting uv as well. I'm not opposed if it benefits us (I'll have to try out the PR to have more of an opinion on that front), that's what ultimately matters the most.

#### Notes

1. (While this is going to sound personal, it's _really_ not. It's simply a recounting of an informative experience I have personally had with the Ruff project.)

    Though it doesn't affect **this** project, one of the challenges for any Python codebase that uses PyGObject is that it's required to violate Ruff's E402 rule against mixing code with imports — GObject-Introspection requires version selection calls prior to module imports, so a typical PyGObject module will look something like this:

    ```python
    import gi
    gi.require_version("GLib", "2.0")
    gi.require_version("Gtk", "4.0")
    from gi.repository import GLib, Gtk
    ```

    Ruff will flag the imports on line 4: "E402 Module level import not at top of file", despite there being no way to write the code so those imports came _before_ the `gi.require_version()` calls. The same thing _would_ be true of imports that come after other, similar statements, like `pytest.importorskip()` calls or `sys.path` modifications — if not for Ruff containing specific logic to recognize and exempt those statements, permitting them to appear before module-level imports without triggering E402.

    Last year I submitted [a Ruff PR](https://github.com/astral-sh/ruff/pull/20289) to add an exception for `gi.require_version()` and `gi.require_versions()` calls, in the same manner. It was rejected (which is fine), on the grounds that they'd rather not add any more exceptions for specific libraries, and instead provide a general configuration-based method of permitting certain code to come before / be interspersed with imports. ...That method did not exist at the time (which is not so fine), and still does not exist today (which is even _less_ fine).

    So, every PyGObject project using Ruff is still forced to `# noqa: E402` all of their `gi.repository` imports. It feels like the perfect is the enemy of the good, a little bit, in the Ruff (and presumably uv) projects.

PR #548, opened 2026-05-21
API: Add context manager for graphs

lkk7:
We can add simple context manager methods for graph classes. 
If you're working in an "imperative" way with non-enormous graphs, this allows for a nice nested workflow.
This doesn't change any existing code.
It doesn't add any management or cleanup, it's just a simple nesting provider.

Example:

```python
from pydot import Cluster, Dot, Edge, Node, Subgraph

with Dot("A") as dot:
    with Subgraph("B") as sg:
        sg.add_node(Node("N1"))
        sg.add_node(Node("N2"))
        sg.add_edge(Edge("N1", "N2"))
        dot.add_subgraph(sg)
    with Cluster("C") as c:
        c.add_node(Node("C1"))
        with Cluster("D") as d:
            d.add_node(Node("D1"))
            c.add_subgraph(d)
        dot.add_subgraph(c)

print(dot.to_string())
dot.write_png("test.png")
```

Results in:

```dot
digraph A {
subgraph B {
N1;
N2;
N1 -> N2;
}
subgraph cluster_C {
C1;
subgraph cluster_D {
D1;
}
}
}
```

test

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -1143,7 +1143,7 @@ class Graph(Common):
         """Exit the runtime context for this graph.
 
         Performs no cleanup.
-        Always returns `False`, so exceptions are never suppressed.
+        Always returns `None`, so exceptions are never suppressed.
         """
         return
 
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -15,7 +15,7 @@ import re
 import subprocess
 import sys
 import warnings
-from typing import Any, Sequence, Union, cast
+from typing import Any, Sequence, TypeVar, Union, cast
 
 import pydot
 from pydot._vendor import tempfile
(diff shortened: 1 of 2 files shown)


src/pydot/core.py
```diff
@@ -943,6 +943,8 @@ def to_string(self, indent: Any = "", indent_level: int = 1) -> str:
 
 __generate_attribute_methods(Edge, EDGE_ATTRIBUTES)
 
+TGraph = TypeVar("TGraph", bound="Graph")
```
lkk7, line 946: This is a way, supported in older Python versions, to return the "self" type in `__enter__`. It would be nice to just do ` -> Self`, but we can't because we want to support older versions with proper handling of subclassing.

lkk7:
Note: this is an experimental feature for now, I'll have to use it in a few cases to improve it and see if it makes sense.

PR #552, opened 2026-06-06
fix: correctly split port from quoted Node ids (closes #489)

### LINKED ISSUE #489: Bug in our Node handling
Discovered while working on #488, because I triggered it in the tests I wrote, but it has nothing to do with the tests or the code they're testing, actually.

```python
>>> from pprint import pp
>>> import pydot
>>> n = pydot.Node("my_name:p45:nw")
>>> pp(n.obj_dict)
{'attributes': {},
 'type': 'node',
 'parent_graph': None,
 'sequence': None,
 'name': 'my_name',
 'port': ':p45:nw'}
>>>
>>> # So far, so good. But...
>>> n2 = pydot.Node('"my_name":p45:nw')
>>> pp(n2.obj_dict)
{'attributes': {},
 'type': 'node',
 'parent_graph': None,
 'sequence': None,
 'name': '"my_name":p45:nw',
 'port': None}
```

...Whoops. So, as a result, those two nodes will have different names, when they _shouldn't_.

AmSach:
Fixes #489.

## Summary
When a Node id was given in the quoted form `"name":port`, pydot
bailed out as soon as it saw the opening quote and stored the entire
string — quotes and all — as the node's name, leaving the port field
as `None`. The unquoted `name:port` form was always split correctly.

The new code finds the matching closing quote (respecting backslash-
escaped quotes inside the name), keeps the surrounding quotes on the
name field so a user-supplied `Node('"node"')` remains distinguish-
able from the DOT reserved keyword `Node('node')`, and extracts any
trailing `:port` as the node's port.

## Before
```
>>> pydot.Node('my_name:p45:nw').get_port()
':p45:nw'   # OK

>>> pydot.Node('"my_name":p45:nw').get_port()
None         # BUG: port swallowed into the name
```

## After
```
>>> pydot.Node('my_name:p45:nw').get_port()
':p45:nw'

>>> pydot.Node('"my_name":p45:nw').get_port()
':p45:nw'   # fixed

>>> pydot.Node('"my_name"').get_port() is None
True        # still works for plain quoted ids
```

## Test
Added a new regression test `test_quoted_node_id_with_port` covering
the bug and the related edge cases (quoted keyword with port, quoted
name containing a colon, unterminated quote). Full suite:
`83 passed, 4 xfailed` — the four xfails are pre-existing
unrelated issues, and the previously-passing 76 tests are unchanged.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -695,13 +695,45 @@ class Node(Common):
             self.obj_dict["parent_graph"] = None
             self.obj_dict["sequence"] = None
 
-            # Remove the compass point
-            #
-            port = None
-            if isinstance(name, str) and not name.startswith('"'):
-                idx = name.find(":")
-                if idx > 0 and idx + 1 < len(name):
-                    name, port = name[:idx], name[idx:]
+            # Split a "<name>"[:port] string into (name, port). The
+            # unquoted "name:port" form has always worked. The quoted
+            # form ('"name":port') used to swallow everything (including
+            # the port) as the name, dropping both the port and the
+            # useful distinction between the two halves of the identifier.
+            # This now finds the matching closing quote (respecting
+            # backslash-escaped quotes inside the name) and treats any
+            # trailing text starting with ":" as a port.
+            port: str | None = None
+            if isinstance(name, str):
+                if name.startswith('"'):
+                    # Find the matching closing quote, respecting \" escapes
+                    closing_idx = -1
+                    i = 1
+                    while i < len(name):
+                        if name[i] == "\\" and i + 1 < len(name):
+                            i += 2
+                            continue
+                        if name[i] == '"':
+                            closing_idx = i
+                            break
+                        i += 1
+                    if closing_idx > 0:
+                        # Capture the part after the closing quote. Keep
+                        # the surrounding quotes on the name so the user's
+                        # explicit "quote it" intent survives a round-trip:
+                        # `Node('"node"')` and `Node('node')` must remain
+                        # distinguishable, since the first is a literal
+                        # node name and the second is a DOT reserved
+                        # keyword. Only the port, if any, is extracted.
+                        rest = name[closing_idx + 1 :]
+                        name = name[: closing_idx + 1]
+                        if rest.startswith(":"):
+                            port = rest
+                else:
+                    idx = name.find(":")
+                    if idx > 0 and idx + 1 < len(name):
+                        port = name[idx:]
+                        name = name[:idx]
 
             if isinstance(name, int):
                 name = str(name)
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -796,6 +796,25 @@ class Node(Common):
 __generate_attribute_methods(Node, NODE_ATTRIBUTES)
 
 
+def _format_node_endpoint(node: "Node | Subgraph | Cluster") -> str:
+    """Return the DOT representation of a node endpoint, including port.
+
+    ``Node`` stores the (optional) port with a leading colon (e.g. ``":p45"``)
+    so the leading colon of the port is treated as part of the node-id syntax.
+    When emitting an Edge endpoint we want ``"name":port`` (or
+    ``"name":"port"`` when the port itself needs quoting), not just the bare
+    name — otherwise the port is silently dropped on round-trip.
+    """
+    name = str(node.get_name())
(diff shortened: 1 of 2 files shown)


src/pydot/core.py
```diff
@@ -695,13 +695,45 @@ def __init__(
             self.obj_dict["parent_graph"] = None
             self.obj_dict["sequence"] = None
 
-            # Remove the compass point
-            #
-            port = None
-            if isinstance(name, str) and not name.startswith('"'):
-                idx = name.find(":")
-                if idx > 0 and idx + 1 < len(name):
```
lkk7: I think removing this check breaks the current behavior

A test file proves it:
```
from pydot import Dot, Node

dot = Dot("A")
dot.add_node(Node('"a"b'))
print(dot.get_nodes()[0].get_name())
print("---")
print(dot.to_string())
```

Before:
```
"a"b
---
digraph A {
"\"a\"b";
}
```

After:
```
"a"
---
digraph A {
"a";
}
```

Not sure if that's what we want.

AmSach:
@lkk7 good catch — fixed in 620126b. The new logic only extracts a port when the text after the closing quote starts with `:`. For input `Node('"a"b')` the rest after the closing quote is `b`, which doesn't start with `:`, so the whole string is now kept as the name (port stays `None`). That matches the pre-fix behavior.

```
name: '"a"b'
port: None
---
digraph A {
"\"a\"b";
}
```

Added a regression test in `test_quoted_node_id_with_port` covering `Node('"a"b')` so this can't regress. Full suite: 278 passed, 4 xfailed. Thanks for the review.

lkk7:
Another interesting difference to notice is:
```
import pydot

n = pydot.Node('"abc":p45')
e = pydot.Edge(n, "b")
print(e.to_string())
```
before: `"abc":p45 -- b;`
after: `"abc" -- b;`

AmSach:
Good catch — fixed in 81e0220. The bug was in Edge.__init__: it only copied the node name into the points tuple, dropping the port that Node had already extracted.

Added `_format_node_endpoint()` that emits `name:port` (or `name:"port"` if quoting is needed) and wired it into both source and destination handling in Edge.__init__.

```
import pydot
n = pydot.Node(abc:p45)
e = pydot.Edge(n, "b")
print(e.to_string())
# before: "abc" -- b;
# after:  "abc":p45 -- b;
```

The round-trip also works: `pydot.Node(e.get_source()).get_port()` is back to `:p45`.

Side effect: the previously-xfailed `test_broken_port_handling` (Node("edge:12") → Edge → str) now passes too, since this is the same code path.

Added a regression test `test_quoted_node_id_with_port_in_edge`. Full suite: 279 passed, 3 xfailed, 1 xpassed (that previously-xfailed one). Thanks again for the review.

AmSach:
@lkk7 Thanks for the reproducer — both regressions are already fixed on the PR head (`81e0220` and `620126b`). Re-running your examples against the current branch:

```python
import pydot

# Case 1 — quoted id with stray trailing text
dot = pydot.Dot("A")
dot.add_node(pydot.Node('"a"b'))
print(dot.get_nodes()[0].get_name())  # -> "a"b
print(dot.to_string())                # -> digraph A { "\""a"b"; }

# Case 2 — quoted id with port used as an Edge endpoint
n = pydot.Node('"abc":p45')
e = pydot.Edge(n, "b")
print(e.to_string())                  # -> "abc":p45 -- b;
```

- `Node('"a"b')` is preserved verbatim (the port extraction only fires when the text after the closing quote starts with `':'`).
- `Edge(pydot.Node('"abc":p45'), 'b')` now serializes as `"abc":p45 -- b;` — the Edge constructor propagates the node's `port` into the endpoint string.

Both cases are locked in by regression tests in `test/test_api.py::test_quoted_node_id_with_port` and `test/test_api.py::test_quoted_node_id_with_port_in_edge`. Full suite is green (`68 passed, 3 xfailed, 1 xpassed`). Mind taking another look?

src/pydot/core.py
ferdnyc: I'm not sure this matters — an ID `"a"b` isn't really valid in Graphviz. As a node name it'll be interpreted as _two_ nodes, `a` and `b`:

```dot
graph G {
    "a"b;
    "a"b -- C;
}
```
produces:
out

ferdnyc:
See #578 for my thoughts about handling `id:port:compass` values in general.

PR #561, opened 2026-07-01
Escape double quotes embedded in node/edge identifiers

binggao1230:
`quote_id_if_necessary` emits a node name or edge endpoint that contains a `"` **raw — unquoted and unescaped** — even though attribute values with the same content are escaped correctly. The result is invalid or silently misparsed DOT:

```pycon
>>> import pydot
>>> g = pydot.Dot("G", graph_type="digraph")
>>> g.add_edge(pydot.Edge('a"b', 'c"d'))
>>> print(g.to_string(), end="")
digraph G {
a"b -> c"d;
}
>>> # reparses to 3 nodes ['a', '"b -> c"', 'd'] and 0 edges — the edge is gone
>>> pydot.graph_from_dot_data('graph G {\na"b;\n}\n')   # -> None (ParseException)
```

A standalone node becomes unparseable, and an edge silently corrupts into three unrelated nodes on round-trip; graphviz `dot` either errors or renders the wrong graph. Node/edge names are arbitrary strings via the public API, so a `"` in one (e.g. `5'2"`, a name from user data) is legal input.

### Cause

`id_needs_quotes('a"b')` returns `False` because `id_re_alpha_nums_with_ports` includes `"` in both of its character classes, so `quote_id_if_necessary` emits the name verbatim instead of routing it through `make_quoted` (which escapes `"` → `\"`).

### Fix

Drop `"` from that regex. An identifier with an embedded `"` then falls through to the port recursion (`id_re_with_port`), which finds no colon to split on, so it is correctly quoted and escaped. Legitimate quoted **ports** are unaffected — they are still recognised by the separate `id_re_with_port` path, so `quote_id_if_necessary('a:"b"')` and `quote_id_if_necessary('a:b')` are both unchanged.

This is a specific, separately-fixable instance of the umbrella issue #187. It is a different code path from the in-flight #552 (which splits ports out of quoted node ids on *input*); this touches the *output* quoting regex only, so the two do not overlap.

`test_quote_id_with_embedded_double_quote` asserts the escaped output, a surviving edge round-trip (1 edge, not 3 nodes), and the preserved port behaviour. The full suite passes (297 passed / 4 xfailed, unchanged from baseline); `ruff check`, `ruff format` and `mypy` are clean.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,7 +9,11 @@ Pydot versions since 1.4.2 adhere to [PEP 440-style semantic versioning]
 4.0.2 (unreleased)
 ------------------
 
-- Nothing changed yet.
+Fixed:
+- Escape double quotes embedded in node and edge identifiers. (#561)
+  Such an identifier was emitted raw and unescaped, producing invalid DOT:
+  a standalone node became unparseable and an edge silently corrupted into
+  three unrelated nodes on round-trip.
 
 
 4.0.1 (2025-06-17)
--- a/test/test_api.py
+++ b/test/test_api.py
@@ -605,6 +605,35 @@ def test_edge_quoting() -> None:
         """)
 
 
+def test_quote_id_with_embedded_double_quote() -> None:
+    """Test that an embedded double quote in an ID is escaped (issue #187).
+
+    A ``"`` embedded in a node name or edge endpoint used to be emitted raw:
+    a standalone node became unparseable and an edge silently corrupted into
+    three unrelated nodes on round-trip.
+    """
+    from pydot.core import quote_id_if_necessary
+
+    # a bare identifier with an embedded quote must be quoted and escaped
+    assert pydot.Node('a"b').to_string() == '"a\\"b";'
+
+    # the edge must survive a round-trip instead of silently corrupting
+    g = pydot.Dot("G", graph_type="digraph")
+    g.add_edge(pydot.Edge('a"b', 'c"d'))
+    assert g.to_string() == textwrap.dedent("""\
+        digraph G {
+        "a\\"b" -> "c\\"d";
+        }
+        """)
+    reparsed = pydot.graph_from_dot_data(g.to_string())
+    assert reparsed is not None
+    assert len(reparsed[0].get_edges()) == 1
+
+    # quoted and plain ports must keep being emitted without extra quoting
+    assert quote_id_if_necessary('a:"b"') == 'a:"b"'
+    assert quote_id_if_necessary("a:b") == "a:b"
+
+
 def test_id_storage_and_lookup() -> None:
     g = pydot.Graph()
     a = pydot.Node("my node")
--- a/test/test_api.py
+++ b/test/test_api.py
@@ -606,32 +606,18 @@ def test_edge_quoting() -> None:
 
 
 def test_quote_id_with_embedded_double_quote() -> None:
-    """Test that an embedded double quote in an ID is escaped (issue #187).
-
-    A ``"`` embedded in a node name or edge endpoint used to be emitted raw:
-    a standalone node became unparseable and an edge silently corrupted into
-    three unrelated nodes on round-trip.
-    """
-    from pydot.core import quote_id_if_necessary
-
-    # a bare identifier with an embedded quote must be quoted and escaped
+    """Test that an embedded double quote in an ID is escaped (issue #187)."""
     assert pydot.Node('a"b').to_string() == '"a\\"b";'
+    assert pydot.Edge("a:b", 'c:"d"').to_string() == 'a:b -- c:"d";'
 
-    # the edge must survive a round-trip instead of silently corrupting
-    g = pydot.Dot("G", graph_type="digraph")
+    g = pydot.Graph("G")
     g.add_edge(pydot.Edge('a"b', 'c"d'))
     assert g.to_string() == textwrap.dedent("""\
         digraph G {
         "a\\"b" -> "c\\"d";
         }
         """)
-    reparsed = pydot.graph_from_dot_data(g.to_string())
-    assert reparsed is not None
-    assert len(reparsed[0].get_edges()) == 1
 
-    # quoted and plain ports must keep being emitted without extra quoting
-    assert quote_id_if_necessary('a:"b"') == 'a:"b"'
-    assert quote_id_if_necessary("a:b") == "a:b"
 
 
 def test_id_storage_and_lookup() -> None:
--- a/test/test_api.py
+++ b/test/test_api.py
@@ -619,6 +619,17 @@ def test_quote_id_with_embedded_double_quote() -> None:
         """)
 
 
+@pytest.mark.xfail(reason="We mishandle two-colon (id:port:compass) endpoints")
+def test_edge_with_port_and_compass() -> None:
+    g = pydot.Graph("G")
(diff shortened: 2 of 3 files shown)


ferdnyc:
I'm telling you, `quote_id_if_necessary` is the bane of my existence.

> Node/edge names are arbitrary strings via the public API, so a `"` in one (e.g. `5'2"`, a name from user data) is legal input.

Well, not so fast there.

The basic statement regarding IDs, with the current logic, is: "If you include quotes, you're in charge of them." I would claim that something like `pydot.Node('a"b')` is invalid input; the user hasn't supplied a properly-quoted ID.

I will probably accept this, because this change does seem to do "better" than what we have now. (Though I have to check some more border cases.)

But it's still possible to screw it up. Your PR code will break on these, same as the `main` branch:

```python
>>> import pydot
>>> g = pydot.Graph("G")
>>> g.add_edge(pydot.Edge('"red"blue"', '<"ab>'))
>>> print(g.to_string(indent=2))
digraph G {
  "red"blue" -> <"ab>;
}
```

There is no way to make the code bulletproof against invalid input.

ChangeLog
```diff
@@ -9,7 +9,11 @@ Pydot versions since 1.4.2 adhere to [PEP 440-style semantic versioning]
 4.0.2 (unreleased)
 ------------------
 
-- Nothing changed yet.
+Fixed:
+- Escape double quotes embedded in node and edge identifiers. (#561)
+  Such an identifier was emitted raw and unescaped, producing invalid DOT:
+  a standalone node became unparseable and an edge silently corrupted into
+  three unrelated nodes on round-trip.
```
ferdnyc: In fact, I have to take issue with this — as I demonstrated, this change at best "slightly improves" the handling of IDs with embedded quotes; it is not a panacea. There were and are valid reasons to be very careful about messing with existing quotes in ID values, and the change here does not in any way give the user free reign to just sprinkle unescaped quotes through their data willy-nilly — it avoids one specific category of breakage with one specific form of (arguably invalid) input.

test/test_api.py
```diff
@@ -649,6 +605,35 @@ def test_edge_quoting() -> None:
...
+    g = pydot.Dot("G", graph_type="digraph")
+    g.add_edge(pydot.Edge('a"b', 'c"d'))
+    assert g.to_string() == textwrap.dedent("""\
+        digraph G {
+        "a\\"b" -> "c\\"d";
+        }
+        """)
+    reparsed = pydot.graph_from_dot_data(g.to_string())
+    assert reparsed is not None
+    assert len(reparsed[0].get_edges()) == 1
```
ferdnyc: I'm not sure the round-trip through graphviz is really necessary here — we know what string forms `dot` will balk at, that isn't in question here. We can check whether we're generating bad strings without running the output back through `dot`.

binggao1230:
Agreed on the semantics. `pydot.Node('a"b')` is pre-quoted input, and pre-quoted IDs remain the caller's responsibility — that is unchanged from `main`, and as your example shows no client-side logic can make invalid input bulletproof.

This change only covers the raw-name case, where a `"` arrives as user data (e.g. `5'2"`) and `main` currently corrupts the output silently. It deliberately does not try to second-guess IDs that already contain quoting.

ferdnyc:
I both looked through the code history and performed further testing.

The double quotes were placed in `id_re_alpha_nums_with_ports` when that expression was originally created, way back in commit ebfd155d60a6000fedb2169c0d963ed5acfcf0e1 (from 2012!).

It seems that in all of my fiddling with that logic, I never touched it except:
1. In ef42633b24797696fa4977f18f947437ecad8380 to change the commas to periods, when we realized that the expressions had been written based on comma decimal separators (some European and South American locales).
2. Again in b358a83892b14b3bf7040bbdc3e1fd48aed6c046, to _remove_ those periods, once we realized that they were valid for numeric IDs, but NOT alphanumeric IDs. (e.g. `1.9` is fine as an ID, `foo.bar` is not unless quoted).

But after a great deal of additional testing, I'm now heavily on the side of merging this. I agree: the double-quote char should never have been part of `id_re_alpha_nums_with_ports` in the first place. It hampers our ability to escape strings containing quotes properly.

The fix _does_ introduce some new...  undesirable behavior, but it's in IDs where things were flat-out broken _without_ this change,

In the `main` branch, this Edge will result in broken graphviz syntax due to the unescaped `"` following the `b` node name:
```python
>>> import pydot
>>> pydot.Edge('a', 'b":5:"sw"').to_string()
'a -- b":5:"sw";'
```

On the PR branch, the IDs now have valid syntax, but the way the escaping and quoting is done becomes... unexpected (and incorrect, from Graphviz's perspective):
```python
>>> import pydot
>>> pydot.Edge('a', 'b":5:"sw"').to_string()
'a -- "b\\":5":"sw";'
```

The quote following the `b` node name gets escaped, but double quotes are wrapped around the _entire_ `b\\":5` portion of the endpoint, which accidentally merges the port into the node name.

It's likely this is just allowing an existing bug in our endpoint handling to manifest in a new way. Pydot doesn't seem to _really_ understand the two-colon endpoint form (<code><var>node_id</var>:<var>port</var>:<var>compass_dir</var></code>) at all.

Yeah, in fact, both the `main` and PR branches do this, which is an existing manifestation of the same problem:

```python
>>> import pydot
>>> pydot.Edge('a', r'foo.bar:5:"sw"').to_string()
'a -- "foo.bar:5":"sw";'
```
It should be `"foo.bar":5:"sw"` (or even `"foo.bar":5:sw`).

So, that's still not any _more_ broken than before, and doesn't make the fix here any less correct.

binggao1230:
Thanks for tracing the history and testing this boundary. I reproduced the same distinction against current `main`: this PR makes `b\":5:\"sw\"` syntactically valid, but Graphviz still folds `:5` into the node ID; `foo.bar:5:\"sw\"` already does the same on `main`. Agreed that two-colon endpoint parsing is a separate existing issue, so I won't widen this PR.

ChangeLog
lkk7: Also, for now we decided not to add stuff to the `ChangeLog`, so you can delete this one to keep it clean. I'll collect all of the changes on next release (maybe it's time to do it soon)

lkk7 (approved):
Thanks, just look at the small nitpicks in the comments (changelog + unnecessary `dot` roundtrip in the test)

ferdnyc:
I'll clean up the test logic here (on the PR branch), since I want to also add an xfail test for that `foo.bar:5:sw` case.

ferdnyc:
Done.

I decided to remove the direct testing of `quote_id_if_necessary`, because while I understand the intent there, that set of unit tests is called `test_api.py` and `quote_id_if_necessary()` isn't _really_ API. (It should probably have been called `_quote_id_if_necessary()`, in retrospect.)

The test function now performs the exact same checks using `pydot.Edge.to_string()`, which employs more of our user-facing API in the way it's meant to be consumed.

ferdnyc:
@lkk7 Honestly I'm ambivalent towards the changelog entry now, double quote handling in the API _was_ more broken than I'd initially realized. Besides, don't you usually rewrite a release's changelog entries just prior to making each release? So... `¯\_(ツ)_/¯`

**Edit:** Oh, I see you mentioned that above. We can just drop that commit from the branch, then. ...Done. This is ready to merge.

ferdnyc (approved):

ChangeLog
ferdnyc: Dropped.

PR #572, opened 2026-08-21
Sync Dot.write and Dot.create signatures

ferdnyc:
The `prog` parameter to `Dot.create` was type-annotated to take any of a list, tuple, str, or None, but `Dot.write` was only typed to accept a str or None. Since `Dot.write()` passed along its `prog` directly to `self.create()` when generating output, clearly they should accept all of the same types.

Redefined both to accept `list[str] | tuple[str] | str | None`.

Also updated both docstrings with more detail about that parameter in particular, and the parameter definitions in general, plus rewrote single-apostrophe fenced literals to use ReST-appropriate double-apostrophe fencing.

I repurposed a test specifically targeting dot calls with arguments supplied (meaning, using a list or tuple `prog=` argument) so that it calls `.write(..., format="svg", prog=[...])` instead of `.write_svg()`. That way, the now-corrected type annotations actually come into play.

We should probably make an effort to reduce use of the monkeypatched convenience functions in our own tests, and rely more on direct calls to `.write()`, `.create()`, `.get()` and `.set()`, since the convenience functions are currently incompatible with static type checking.

(Looking over the test code again, we're actually pretty good about that. For the most part, generated methods aren't used except in the few tests where we're explicitly _testing_ the generated methods.)

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -1755,22 +1755,30 @@ class Dot(Graph):
     def write(
         self,
         path: str | bytes,
-        prog: str | None = None,
+        prog: list[str] | tuple[str] | str | None = None,
         format: str = "raw",
         encoding: str | None = None,
     ) -> bool:
         """Writes a graph to a file.
 
-        Given a filename 'path' it will open/create and truncate
-        that file and write to it a representation of the graph
-        defined by the dot object in the format specified by
-        'format' and using the encoding specified by `encoding` for text.
-        The format 'raw' is used to dump the string representation
-        of the Dot object, without further processing.
-        The output can be processed by any of graphviz tools, defined
-        in 'prog', which defaults to 'dot'
-        Returns True or False according to the success of the write
-        operation.
+        Creates/opens (and truncates) the file ``path``, and writes
+        to it a rendered representation of the current graph.
+
+        The file format can be specified with ``format``, which defaults
+        to ``raw`` (see below).
+
+        The encoding to use when writing the intermediate graph
+        definition can be set with ``encoding``. (The default is
+        ``utf-8``, which matches the data in memory and is almost
+        always what you want.)
+
+        The special ``format`` value ``raw`` is used to dump the generated
+        graph definition ("dotfile" source), without further processing.
+        (In ``raw`` mode, ``encoding`` can be used to set the encoding of
+        the output file.)
+
+        Returns True, or raises an exception if an error is encountered.
+        The exception is some form of ``OSError`` or an ``AssertionError``.
 
         There's also the preferred possibility of using:
 
@@ -1779,9 +1787,45 @@ class Dot(Graph):
         which are automatically defined for all the supported formats.
         [write_ps(), write_gif(), write_dia(), ...]
 
-        The encoding is passed to `open` [1].
+        The encoding is passed to ``open`` [1].
 
-        [1] https://docs.python.org/3/library/functions.html#open
+        [1]: https://docs.python.org/3/library/functions.html#open
+
+        @param prog:
+
+          One of:
+
+          - the name of Graphviz executable that
+            can be found in the ``$PATH``.
+
+          - the absolute path to Graphviz executable.
+
+          - a sequence (list, tuple) with the first
+            element being one of the above, and the
+            remaining elements being command-line
+            arguments to use when invoking the executable.
+
+          If you have added Graphviz to the ``$PATH`` and
+          use its executables as installed
+          (without renaming any of them)
+          then their names are:
+
+            - ``dot``
+            - ``twopi``
+            - ``neato``
+            - ``circo``
+            - ``fdp``
+            - ``sfdp``
+
+          On Windows, these have the notorious ".exe" extension that,
+          only for the above strings, will be added automatically.
+
+          The ``$PATH`` is inherited from ``os.env['PATH']`` and
+          passed to ``subprocess.Popen`` using the `env` argument.
+
+          If you haven't added Graphviz to your ``$PATH`` on Windows,
+          then you may want to give the absolute path to the
+          executable (for example, to ``dot.exe``) in ``prog``.
         """
         if prog is None:
             prog = self.prog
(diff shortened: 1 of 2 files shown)


src/pydot/core.py
```diff
@@ -1885,22 +1933,22 @@ def create(
...
                     working_dir=tmp_dir,
                 )
             except OSError as e:
                 if e.errno == errno.ENOENT:
-                    args = list(e.args)  # type: ignore
-                    args[1] = f'"{prog}" not found in path.'  # type: ignore
-                    raise OSError(*args)
+                    errargs: list[str] = list(e.args)
+                    errargs[1] = f'"{_prog}" not found in path.'
+                    raise OSError(*errargs)
```
ferdnyc, line 1959: Always nice when eliminating variable redefinition lets you dump a bunch of `# type: ignore` comments.

ferdnyc:
ARRRRRGH.

Stupid mistake on my part, but at the same time feels like I fell into a trap. It's kind of insane that `str` is a `Sequence` (it's an `Iterable`, too):

```python3
>>> from typing import Sequence
>>> s = "red"
>>> isinstance(s, Sequence)
True
```

...Which means, not only was my `Sequence[str] | str | None` type-annotation actually **redundant** (`Sequence[str]` also encompasses `str` itself), but I can't use `isinstance(prog, Sequence)` to differentiate between a list/tuple `prog` and a string `prog`. It'll be true for all three. Grrr.

**Edit:** Fixed to use `list[str] | tuple[str] | str | None` again. That's what I get for trying to be clever.

**Edit2:** Similarly insane:

```python
>>> a_bool = True
>>> isinstance(a_bool, int)
True
```

src/pydot/core.py
```diff
@@ -1755,20 +1755,22 @@ def set_prog(self, prog: str) -> None:
...
+        to it a rendered representation of the current graph.
+        The file format can be specified with ``format``, and the encoding
+        to use when ``open()``-ing the file can be set with ``encoding``.
+        (``encoding`` is only meaningful for text-based output formats,
+        such as ``ps`` or ``svg``, not binary image formats.)
+
+        The special ``format`` value ``raw`` is used to dump the generated
+        graph definition ("dotfile" source), without further processing.
+
         Returns True or False according to the success of the write
```
lkk7: I know you didn't touch that line, but while we're at it, I think it either returns True or fails. Nothing else possible

src/pydot/core.py
```diff
@@ -1755,20 +1755,22 @@ def set_prog(self, prog: str) -> None:
...
-        'format' and using the encoding specified by `encoding` for text.
-        The format 'raw' is used to dump the string representation
-        of the Dot object, without further processing.
-        The output can be processed by any of graphviz tools, defined
-        in 'prog', which defaults to 'dot'
+        Creates/opens (and truncates) the file ``path``, and writes
+        to it a rendered representation of the current graph.
+        The file format can be specified with ``format``, and the encoding
+        to use when ``open()``-ing the file can be set with ``encoding``.
+        (``encoding`` is only meaningful for text-based output formats,
```
lkk7: I'm not sure if `encoding` is only meaningful for text-based outputs. Both for `write` and `create`:

```
import pydot

graph = pydot.Dot(graph_type="digraph")
graph.add_node(pydot.Node("a", label="Łódź"))

# Works:
png = graph.create_png(encoding="utf-8")

# Fails
graph.create_png(encoding="ascii")
```

ferdnyc: That's fair, maybe "useful" is a better word than "meaningful". The only thing you can do by passing `encoding` for binary outputs is break them.

ferdnyc: Oh, I see what you're saying. But that's not because of the graph syntax encoding, is it? I would expect that this would also fail, as PNG can't be encoded in ASCII:

```python
g = pydot.Dot("G")
g.add_node(pydot.Node("Fred"))
g.write_png("foo.png", encoding="ascii")
```

Unless I'm misunderstanding where `encoding` is used in the output code.

ferdnyc: Nope, I'm wrong. The above writes a perfectly readable PNG; `encoding` is only for the graphviz syntax. So, yeah, I'll fix those descriptions.

ferdnyc: (I kind of hate that `encoding` is an output parameter, then. If a graph has _data_ that uses a particular encoding, then that should really be a property of the graph definition itself, set on creation or with a `pydot.Graph.set_encoding()` call. IMHO.)

src/pydot/core.py
ferdnyc: Oy. It's even slightly worse than that, really — `write()` _always_ returns `True`, since it doesn't even have any failure logic.

Technically you're correct, but the "fails" part of those calls is so poorly-defined it's hard to even claim it's an intentional part of the program logic.

If graph creation fails in `.create()` (and doesn't raise an exception trying to execute the subprocess, which can happen) then it's due to a non-zero `dot` exit code, in which case the `.create()` call will:
  1. print an error message to stdout
  2. fail an `assert` that the subprocess exited successfully (Which is a _bizarre_ form of error handling for an API. If you're going to rely on an `AssertionError` exception being raised, why not at least raise a custom one with some useful information? It's not like we don't have our own Exception classes defined.)

And then, if it hadn't been derailed by the `AssertionError` raised by `.create()` — which it does not catch or attempt to manage in any way — `.write()` would blindly `True` no matter what!

src/pydot/core.py
ferdnyc: Come to think of it, if `encoding` is only used on output, then it's kind of insane that `pydot.Graph.to_string()` doesn't also have an `encoding` parameter. Or at least a `.to_bytes()`/`.to_encoded_string()` that takes an `encoding` and returns encoded `bytes`.

It feels like a misunderstanding of how character encoding works, that Pydot assumes it only matters if you're writing the graph definition out to a file.

src/pydot/core.py
ferdnyc: I updated both descriptions to indicate that they return True or bytes (as appropriate), or raise either `OSError` or `AssertionError`.

src/pydot/core.py
ferdnyc: Anyway, descriptions updated to explain what `encoding` actually does, and why you almost certainly don't want to change it unless you're using `.write()` with `format=raw`.

src/pydot/core.py
ferdnyc: Whoops, missed one.

```suggestion
            f'"{_prog}" with args {arguments} '
```

PR #573, opened 2026-08-21
Accept Node/Subgraph objects in get_node() and get_edge()

### LINKED ISSUE #272: get_edge() does not work with node instances as imput
```
import pydot

g = pydot.Dot()
n1 = pydot.Node('n1')
n2 = pydot.Node('n2')
g.add_node(n1)
g.add_node(n2)
g.add_edge(pydot.Edge(src=n1, dst=n2))

print(g.get_edge(n1, n2))
```

return an empty list.

inahus99:
Closes #272

### The problem

`Edge` reduces its endpoints to names on construction, and `add_node` keys nodes by name. The lookups never applied the same reduction, so passing the objects back returned nothing:

```python
import pydot

g = pydot.Dot()
n1 = pydot.Node('n1')
n2 = pydot.Node('n2')
g.add_node(n1)
g.add_node(n2)
g.add_edge(pydot.Edge(src=n1, dst=n2))

print(g.get_edge(n1, n2))  # [] -- expected the edge
print(g.get_node(n1))      # [] -- same cause
```

### The change

`get_node()` and `get_edge()` now resolve `Node`/`Subgraph`/`Cluster` arguments to their names before looking them up.

This is only bringing the getters in line with the deleters: `del_node()` and `del_edge()` already did exactly this, inline. Those inline `isinstance` checks are now the shared `name_of()` helper, so there's one place that defines what an endpoint reduces to.

`Cluster` is a `Subgraph` subclass as of #571, so the isinstance check covers all three types.

### Notes on scope

- Anything that isn't a `Node`/`Subgraph`/`Cluster` is passed through untouched, so names, ints, floats and `FrozenDict` endpoints keep working as keys exactly as before.
- Lookup by name is unchanged, and the multi-element list form of `get_edge()` keeps its existing behaviour.
- No change to what `get_edge()` returns: it still hands back the edge as stored, so a reversed lookup on an undirected graph returns the original orientation. There's a test pinning that.

### Testing

Six tests added, covering lookup by `Node`, by the list form, by `Subgraph`/`Cluster`, the undirected reversed case, that name lookups still work, and that `del_edge` and `get_edge` now agree.

`ruff check`, `ruff format --check` and `mypy` (strict) are clean. I couldn't install Graphviz on this machine, so `TestGraphvizRegressions` and three other tests that shell out to `dot` don't run locally; the rest of the suite passes, and none of it touches the subprocess path.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -958,6 +958,21 @@ class Edge(Common):
 __generate_attribute_methods(Edge, EDGE_ATTRIBUTES)
 
 
+def name_of(obj: Any) -> Any:
+    """Return the name of a `Node`, `Subgraph` or `Cluster`.
+
+    Any other value is returned unchanged, so that names, numbers and
+    `FrozenDict` endpoints keep working as lookup keys.
+
+    Nodes and edges are stored under their name, because `Edge` reduces
+    its endpoints to names on construction. Lookups therefore have to
+    apply the same reduction to find them again.
+    """
+    if isinstance(obj, (Node, Subgraph)):
+        return str(obj.get_name())
+    return obj
+
+
 class Graph(Common):
     """Class representing a graph in Graphviz's dot language.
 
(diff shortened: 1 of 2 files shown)


lkk7:
Upstream issue has been closed.

Issue #220, opened 2019-10-07
RFE: Rendet Dot results in Jupyter cells


ankostis:
Simply adding this method in `pydot.Dot` class makes it renderable in jupyter notebooks & lab:

```diff
diff --git a/pydot.py b/pydot.py
index 60a2bac..70829b6 100644
--- a/pydot.py
+++ b/pydot.py
@@ -1745,6 +1745,8 @@ class Dot(Graph):
 
         self.obj_dict = state
 
+    def _repr_svg_(self):
+        return self.create_svg().decode()
 
     def set_shape_files(self, file_paths):
         """Add the paths of the required image files.
```

There is an issue of which encoding to use, and how.
But it would be a great facility to data people who love notebooks.

ankostis:
Note that `_repr_svg_()` has some drawbacks on _jupyterlab_ (but not plain  Jupyter notebooks,, as described in jupyterlab/jupyterlab#7497.

ankostis:
(continuing my last comment)
I reverted to implement the `_repr_html()` method instead, which works in both jupyter environments, but the SVGs are not re-scalable.  Therefore i used the [`svg-pan-zoom` JS library](https://github.com/ariutta/svg-pan-zoom/) to make them "zoomable" with user controls, as shown in this hackish commit: pygraphkit/graphtik@e350afa78.

For the web, i would try also the all client-JS solution https://visjs.github.io/vis-network/docs/network

ankostis:
For future reference, this the code for zoomable SVGs in Jupyter:

```python
def _repr_html_monkeypatches(dot):
    """
    Monkey-patching for ``pydot.Dot._repr_html_()` for rendering in jupyter cells.

    :param dot:
        a `pydot.Dot()` instance

    .. Note::
        Had to use ``_repr_html_()`` and not simply ``_repr_svg_()`` because
        (due to https://github.com/jupyterlab/jupyterlab/issues/7497)

    """
    pan_zoom_json" = "{controlIconsEnabled: true, zoomScaleSensitivity: 0.4, fit: true}",
    element_styles = "width: 100%; height: 300px;",
    container_styles = ""
    svg_txt = dot.create_svg().decode()
    html = f"""
        <div class="svg_container">
            <style>
                .svg_container {{
                    {container_styles}
                }}
                .svg_container SVG {{
                    {element_styles}
                }}
            </style>
[...]

PR #574, opened 2026-08-21
Docs: document rendering a graph in a Jupyter notebook

### LINKED ISSUE #286: Document SVG Rendering for Jupyter Notebook
In Jupyter Notebook, this code generates an SVG from `graph`, and display the rendering below the cell.
```python
from IPython.display import SVG

SVG(graph.create_svg())
```

It would be nice if this were documented in the README.md file.

This code seems to be a way to use pydot in jupyter notebook and get the interactivity of the environment.  Issue #220 seems to take this to the next level.

Thank you for the library!

inahus99:
Closes #286

Adds a short "Display in a Jupyter notebook" subsection under *3. Output* in the README:

```python
from IPython.display import SVG

SVG(graph.create_svg())
```

The `create_*` methods already return exactly what the IPython display classes want, but nothing in the docs said so, so it wasn't discoverable. The section also mentions `Image(graph.create_png())` for the bitmap case, and notes that both go through Graphviz.

Verified the snippets against IPython 9.16 — `SVG()` accepts the `bytes` that `create_svg()` returns, no decode needed.

Deliberately scoped to documenting what already works. #220 asks for a `_repr_svg_` on the graph objects so a bare `graph` renders in a cell; that's a code change and a separate issue, so it isn't touched here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,8 +202,7 @@ There are two different DOT strings you can retrieve:
 
 #### Display in a Jupyter notebook
 
-Pass the output of a `create_*` method to the matching IPython display
-class to render the graph below a notebook cell:
+To render the graph below a notebook cell:
 
 ```python
 from IPython.display import SVG
@@ -211,16 +210,13 @@ from IPython.display import SVG
 SVG(graph.create_svg())
 ```
 
-SVG keeps the graph sharp at any zoom level and keeps the text
-selectable. For a bitmap instead, use `Image` with `create_png()`:
+Or, as a bitmap:
 
 ```python
 from IPython.display import Image
 
 Image(graph.create_png())
 ```
-
-Both call Graphviz, so they render the graph as Graphviz lays it out.
 
 #### Convert to a NetworkX graph
 
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ graph.get_node("b")[0].set_shape("box")
 
 ### 3. Output
 
-Here are three different output options:
+Here are four different output options:
 
 #### Generate an image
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -200,6 +200,28 @@ There are two different DOT strings you can retrieve:
   graph.write_dot("output_graphviz.dot")
   ```
 
+#### Display in a Jupyter notebook
+
+Pass the output of a `create_*` method to the matching IPython display
```
lkk7: No need to mention `create_*`, there are only two options in the docs here. Maybe just "To render the graph below a notebook cell:" is enough

README.md
```diff
@@ -200,6 +200,28 @@ There are two different DOT strings you can retrieve:
...
+class to render the graph below a notebook cell:
+
+```python
+from IPython.display import SVG
+
+SVG(graph.create_svg())
+```
+
+SVG keeps the graph sharp at any zoom level and keeps the text
```
lkk7: No need to explain SVG and PNG characteristics, we can keep it short.

README.md
```diff
@@ -200,6 +200,28 @@ There are two different DOT strings you can retrieve:
...
+selectable. For a bitmap instead, use `Image` with `create_png()`:
+
+```python
+from IPython.display import Image
+
+Image(graph.create_png())
+```
+
+Both call Graphviz, so they render the graph as Graphviz lays it out.
```
lkk7: This line can be omitted, in general Graphviz is needed for the whole pydot stuff

lkk7 (changes requested):
The code is good, but in my opinion the code explanations are too verbose.

README.md
inahus99: Done now just "To render the graph below a notebook cell:".

README.md
inahus99: Done cut the SVG/PNG comparison. The second snippet is introduced with just "Or, as a bitmap:".

README.md
inahus99: Done, removed. Agreed  Graphviz is already covered under Dependencies.

inahus99:
Fair point trimmed in a541505. The section is now two snippets with a one-line lead each, and the prose is down from five sentences to two fragments:

```markdown
#### Display in a Jupyter notebook

To render the graph below a notebook cell:

```python
from IPython.display import SVG

SVG(graph.create_svg())
```

Or, as a bitmap:

```python
from IPython.display import Image

Image(graph.create_png())
```
```

Left both formats in, since your first comment read as "don't call them `create_*`" rather than "drop one" .

`reuse lint` passes locally (the other two workflows skip this branch on `paths-ignore`).

lkk7 (approved):

Issue #553, opened 2026-06-22
Bug: `EDGE` and `NODE` not properly quoted when used as nodes


bertptrs:
`EDGE` and `NODE` are keywords in dot, and so if you have a node named NODE, you have to quote it to refer to it. `pydot` correctly escapes these nodes when using them in edges, but not in the original node definition. To illustrate, the following script:

```python
import pydot

graph = pydot.Dot("my_graph", graph_type="digraph")
node_node = pydot.Node("NODE")
node_edge = pydot.Node("EDGE")
graph.add_node(node_node)
graph.add_node(node_edge)

edge = pydot.Edge("NODE", "EDGE")
graph.add_edge(edge)
print(graph.to_string())
```

results in the output

```
digraph my_graph {
NODE;
EDGE;
"NODE" -> "EDGE";
}
```

which is rejected by the common dot tools. Instead, the correct output would've been:

```
digraph my_graph {
"NODE";
"EDGE";
"NODE" -> "EDGE";
}
```

lkk7:
Thanks!

There is a bug, but a different one...

An unquoted keyword passed to `Node()`, like `Node("node")` or `Node("NODE")`, should be treated as a default-attribute setter statement `(node [attrs...])`, e.g. "set color on all nodes to red", not as a node identifier. That's why `Node("node")` with no attributes produces no output, because bare `node;` with no attributes is meaningless in DOT.

The actual bug is that `Node("NODE")` isn't being treated consistently with `Node("node")`. I'll fix it soon.

For your use case, creating an actual `"NODE"` node,  you need to embed the quotes in the name string:
```
node_node = pydot.Node('"NODE"')
```

Now, I know this may be confusing, but... I'm just a maintainer :)

bertptrs:
Interesting, thank you for getting back to me.

I originally hit this while using [`networkx::write_dot`](https://networkx.org/documentation/stable/reference/generated/networkx.drawing.nx_pydot.write_dot.html) and since I could reproduce it using this library by itself, I mistakenly thought this was the bug. Now there are _two_ bugs.

Is there a way to safely add a node to the graph, regardless of whether it's a keyword or not? Because either:

- `networkx` will need to know what keywords exist in `dot` while writing the graph, or
- `networkx` should switch to the "safe" method of writing nodes

I guess it's safe to _always_ quote if necessary.

lkk7:
You can try the `quote_id_if_necessary` function that's used internally. 

[...]

PR #575, opened 2026-08-21
Don't emit a blank line for a suppressed default statement

inahus99:
Refs #553

### What this is (and isn't)

This is **not** a proposal to change keyword handling. In #553 you were clear that an unquoted keyword passed to `Node()` is a default-attribute statement, that `Node("NODE")` should behave like `Node("node")`, and that `Node('"NODE"')` is how you ask for a literal node — and #554 implemented that. All of that is left exactly as it is, and the two `..._uppercase_...` tests from #554 still pass unmodified.

This only cleans up the leftover whitespace from the suppression path.

### The problem

`Node.to_string()` returns `""` for a default statement that sets no attributes, since bare `node;` is meaningless. `Graph.to_string()` appended a newline for it regardless, so every suppressed statement left a blank line:

```python
g = pydot.Dot("my_graph", graph_type="digraph")
g.add_node(pydot.Node("NODE"))
g.add_node(pydot.Node("EDGE"))
g.add_edge(pydot.Edge("NODE", "EDGE"))
print(repr(g.to_string()))
```

Before:
```
'digraph my_graph {\n\n\n"NODE" -> "EDGE";\n}\n'
```

After:
```
'digraph my_graph {\n"NODE" -> "EDGE";\n}\n'
```

Worth noting the output in the original #553 report is already valid DOT on `main` — #554 fixed that part. These blank lines are the only visible remainder of that example.

### The change

Skip the append when the node renders empty, instead of appending a lone `"\n"`.

A default statement that *does* set attributes is still emitted, and which names count as keywords is untouched. Two tests: one for the blank lines, one pinning that `set_node_defaults()` output still renders.

`ruff check`, `ruff format --check` and `mypy` (strict) are clean. Graphviz isn't installed on this machine, so `TestGraphvizRegressions` doesn't run locally — that suite compares rendered output, and this changes only whitespace between statements, but it's worth a look at CI.

Happy to close this if you'd rather keep #553 as a single thread.

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -1546,6 +1546,13 @@ class Graph(Common):
                 node_str = node.to_string(
                     indent=indent, indent_level=indent_level + 1
                 )
+
+                # A default-attribute statement that sets no attributes
+                # renders as the empty string. Emitting a newline for it
+                # would leave a blank line behind.
+                if not node_str:
+                    continue
+
                 graph.append(f"{node_str}\n")
 
             elif obj["type"] == "edge":
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -1547,9 +1547,7 @@ class Graph(Common):
                     indent=indent, indent_level=indent_level + 1
                 )
 
-                # A default-attribute statement that sets no attributes
-                # renders as the empty string. Emitting a newline for it
-                # would leave a blank line behind.
+                # May be empty, avoid blank line
                 if not node_str:
                     continue
 
--- a/test/test_api.py
+++ b/test/test_api.py
@@ -388,7 +388,7 @@ def test_suppressed_default_stmt_leaves_no_blank_line() -> None:
     g.add_node(pydot.Node("NODE"))
     g.add_node(pydot.Node("EDGE"))
     g.add_edge(pydot.Edge("NODE", "EDGE"))
-    assert g.to_string() == ('digraph testgraph {\n"NODE" -> "EDGE";\n}\n')
+    assert g.to_string() == 'digraph testgraph {\n"NODE" -> "EDGE";\n}\n'
 
 
 def test_default_stmt_with_attributes_still_emitted() -> None:
@@ -396,7 +396,7 @@ def test_default_stmt_with_attributes_still_emitted() -> None:
     g.set_node_defaults(shape="box")
     g.add_node(pydot.Node("a"))
     g.add_node(pydot.Node("node"))
-    assert g.to_string() == ("digraph testgraph {\nnode [shape=box];\na;\n}\n")
+    assert g.to_string() == "digraph testgraph {\nnode [shape=box];\na;\n}\n"
 
 
 def test_dot_graph_type_uppercase() -> None:
--- a/test/test_api.py
+++ b/test/test_api.py
@@ -383,6 +383,22 @@ def test_keyword_node_id_uppercase_to_string_with_attributes() -> None:
     assert g.get_nodes()[0].to_string() == "NODE [shape=box];"
 
 
+def test_suppressed_default_stmt_leaves_no_blank_line() -> None:
+    g = pydot.Dot("testgraph", graph_type="digraph")
+    g.add_node(pydot.Node("NODE"))
+    g.add_node(pydot.Node("EDGE"))
+    g.add_edge(pydot.Edge("NODE", "EDGE"))
+    assert g.to_string() == ('digraph testgraph {\n"NODE" -> "EDGE";\n}\n')
+
+
+def test_default_stmt_with_attributes_still_emitted() -> None:
+    g = pydot.Dot("testgraph", graph_type="digraph")
+    g.set_node_defaults(shape="box")
+    g.add_node(pydot.Node("a"))
+    g.add_node(pydot.Node("node"))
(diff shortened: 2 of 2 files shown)


src/pydot/core.py
```diff
@@ -1546,6 +1546,13 @@ def to_string(
                 node_str = node.to_string(
                     indent=indent, indent_level=indent_level + 1
                 )
+
+                # A default-attribute statement that sets no attributes
+                # renders as the empty string. Emitting a newline for it
+                # would leave a blank line behind.
+                if not node_str:
+                    continue
+
```
ferdnyc, line 1553: The _change_ is fine — good catch, thanks!

But, please trim the comment down to bare necessities only. The source code shouldn't read like a development journal.

```suggestion
                # May be empty, avoid blank line
                if not node_str:
                    continue
```

test/test_api.py
```diff
@@ -383,6 +383,22 @@ def test_keyword_node_id_uppercase_to_string_with_attributes() -> None:
     assert g.get_nodes()[0].to_string() == "NODE [shape=box];"
 
 
+def test_suppressed_default_stmt_leaves_no_blank_line() -> None:
+    g = pydot.Dot("testgraph", graph_type="digraph")
+    g.add_node(pydot.Node("NODE"))
+    g.add_node(pydot.Node("EDGE"))
+    g.add_edge(pydot.Edge("NODE", "EDGE"))
+    assert g.to_string() == ('digraph testgraph {\n"NODE" -> "EDGE";\n}\n')
```
ferdnyc: Why are the expected strings in parentheses? It makes them read confusingly like tuples.
```suggestion
    assert g.to_string() == 'digraph testgraph {\n"NODE" -> "EDGE";\n}\n'
```

test/test_api.py
```diff
@@ -383,6 +383,22 @@ def test_keyword_node_id_uppercase_to_string_with_attributes() -> None:
...
+
+
+def test_default_stmt_with_attributes_still_emitted() -> None:
+    g = pydot.Dot("testgraph", graph_type="digraph")
+    g.set_node_defaults(shape="box")
+    g.add_node(pydot.Node("a"))
+    g.add_node(pydot.Node("node"))
+    assert g.to_string() == ("digraph testgraph {\nnode [shape=box];\na;\n}\n")
```
ferdnyc:
```suggestion
    assert g.to_string() == "digraph testgraph {\nnode [shape=box];\na;\n}\n"
```

ferdnyc (changes requested):
The logic is sound, and tests are included which is always appreciated. Some style adjustments are all that's needed.

src/pydot/core.py
ferdnyc: (Note: A **slightly** more verbose comment would also be fine, as long as it's still cut down to only necessary detail. These are all reasonable alternatives:)

* `# node_str may be empty`
* `# node_str may be empty, suppress blank line`
* `# May be empty, avoid blank line in output`
* `# node_str may be empty, don't output blank line`
* `# Don't output blank line on empty string`

inahus99: Agreed. Took your wording verbatim: `# May be empty, avoid blank line`. Point taken on the journal thing.

test/test_api.py
inahus99: Mine, not ruff's. I'd wrapped the assert while it was longer, and the parens outlived the reason for them. Both removed; the lines are 73 and 77 chars, and `ruff format --check` is happy without them.

test/test_api.py
inahus99: Same here, removed.

src/pydot/core.py
inahus99: Thanks for the follow-up. Went with your original one-liner, which reads fine at the call site.

inahus99:
All four addressed in a8cfe42. Thanks for the quick and specific review @ferdnyc.

- Comment trimmed to your suggested `# May be empty, avoid blank line`.
- Parentheses dropped from both expected strings. Those were mine rather than the formatter's; I'd wrapped the asserts while they were longer and the parens stuck around after. They're 73 and 77 chars now, and `ruff format --check` leaves them alone.

Locally: `ruff check`, `ruff format --check`, `mypy` (strict) and `reuse lint` all clean, and the suite is at 103 passed. The four failures I see are `dot`-not-on-PATH. I couldn't get Graphviz installed on this machine, so `TestGraphvizRegressions` still hasn't run anywhere. Since this touches DOT text assembly, that suite is the one worth having CI confirm.

@lkk7 this one's still gated on workflow approval for a first-time contributor, so no checks have run on any of the three PRs yet. Whenever you have a moment.

lkk7:
@inahus99 

>  I couldn't get Graphviz installed on this machine

On what machine? 😉 Is this Copilot in cloud or what?
Are you implying you're making changes to pydot without having access to Graphviz?

inahus99:
@lkk7 haha , on my lenovo slim 5 ,gen 10, amd r7:)

inahus99:
@lkk7  since the change only touched the DOT string assembly logic, I initially just verified it against the unit tests and linters in Python before setting up the external binaries.

Got Graphviz installed on Windows now and mapped to PATH. Ran the full regression suite locally and all 107 tests are passing cleanly.
You can merge it without any issues

ferdnyc (approved):

lkk7 (approved):

PR #577, opened 2026-08-26
Type-checking for our tests

ferdnyc:
A whole bunch of commits in service of the final one: Add the
`test/` directory to the list of locations MyPy will type-check.

To make that work, a large number of adjustments had to be made to the test scripts, and a few to the core classes as well. Some of the core fixes represent API changes. Like:

* The `pydot.core.Node()` constructor accepts `int`, `float,` and `bool` as its `name=` argument, in addition to `str`. Whatever type the parameter is, it stores it in `obj_dict` as `str(name)`. (The latter was existing code.) 
    Necessary because some of our own tests are calling `pydot.Node(1)` or `pydot.Node(True)` and expecting that to work.
* `pydot.core.Node.set_name()` now also accepts `int`, `float`, and `bool` in addition to `str`, and does NOT accept `None`. For parity with the constructor, it now also stores `str(node_name)` in the `obj_dict`.
* `pydot.core.Common.set_parent_graph()` used to accept `Common` as its argument, a consequence of `Graph` not being defined yet. Now that we're using `from __future__ import annotations`, that's not a problem and it can be updated to the correct `Graph`.

Other changes are non-API. 

* Lots of typing fixes in the tests. All of the calls to generated methods have to be `# type: ignore`d until we solve the problem of dynamic API not being properly type-checkable.
* Some tweaks to the code layout in `pydot.core.Graph` and the module imports (no functional changes)
* Use aliased explicit imports (`from pydot.exceptions import Error as Error`) in `__init__`, as a supported method of evading Ruff's "F401 unused-import".

### DIFF
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -577,7 +577,7 @@ class Common:
     def __setstate__(self, state: AttributeDict) -> None:
         self.obj_dict = state
 
-    def set_parent_graph(self, parent_graph: Common | None) -> None:
+    def set_parent_graph(self, parent_graph: Graph | None) -> None:
         self.obj_dict["parent_graph"] = parent_graph
 
     def get_parent_graph(self) -> Graph | None:
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -690,7 +690,7 @@ class Node(Common):
 
     def __init__(
         self,
-        name: str = "",
+        name: str | int | float | bool = "",
         obj_dict: AttributeDict | None = None,
         **attrs: Any,
     ) -> None:
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -17,14 +17,15 @@ import sys
 import warnings
 from typing import TYPE_CHECKING, Any, Final, Sequence, Union, cast
 
-if TYPE_CHECKING:
-    # `typing_extensions` is always available in `TYPE_CHECKING` blocks,
-    # even if not  installed
-    from typing_extensions import Self, TypeAlias
-
 import pydot
 from pydot._vendor import tempfile
 from pydot.classes import AttributeDict, EdgeEndpoint, FrozenDict
+
+if TYPE_CHECKING:
+    # `typing_extensions` is always available in `TYPE_CHECKING` blocks,
+    # even if not installed
+    from typing_extensions import Self, TypeAlias
+
 
 _logger = logging.getLogger(__name__)
 _logger.debug("pydot core module initializing")
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -1041,6 +1041,28 @@ class Graph(Common):
     def __str__(self) -> str:
         return self.to_string()
 
+    def __enter__(self) -> Self:
+        """Enter the runtime context for this graph.
+
+        Enables a nested construction style:
+
+            with Dot("A") as dot:
+                with Subgraph("B") as sg:
+                    sg.add_node(Node("N1"))
+                    dot.add_subgraph(sg)
+
+        This is syntactic sugar only with no side effects.
+        """
+        return self
+
+    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
+        """Exit the runtime context for this graph.
+
+        Performs no cleanup.
+        Always returns `None`, so exceptions are never suppressed.
+        """
+        return
+
     def get_graph_type(self) -> str | None:
         return self.obj_dict["type"]  # type: ignore
 
@@ -1146,28 +1168,6 @@ class Graph(Common):
         seq: int = self.obj_dict.get("current_child_sequence", 1)
         self.obj_dict["current_child_sequence"] = seq + 1
         return seq
-
-    def __enter__(self) -> Self:
-        """Enter the runtime context for this graph.
-
-        Enables a nested construction style:
-
-            with Dot("A") as dot:
-                with Subgraph("B") as sg:
-                    sg.add_node(Node("N1"))
-                    dot.add_subgraph(sg)
-
-        This is syntactic sugar only with no side effects.
-        """
-        return self
-
-    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
-        """Exit the runtime context for this graph.
-
-        Performs no cleanup.
-        Always returns `None`, so exceptions are never suppressed.
-        """
-        return
 
     def add_node(self, graph_node: Node) -> None:
         """Adds a node object to the graph.
--- a/src/pydot/core.py
+++ b/src/pydot/core.py
@@ -720,13 +720,13 @@ class Node(Common):
     def __str__(self) -> str:
         return self.to_string()
 
-    def set_name(self, node_name: str | None) -> None:
+    def set_name(self, node_name: str | int | float | bool) -> None:
         """Set the node's name."""
-        self.obj_dict["name"] = node_name
+        self.obj_dict["name"] = str(node_name)
 
     def get_name(self) -> str:
         """Get the node's name."""
-        return self.obj_dict["name"]  # type: ignore
+        return str(self.obj_dict["name"])
 
     def get_port(self) -> str | None:
         """Get the node's port."""
--- a/test/test_pydot.py
+++ b/test/test_pydot.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import functools
 import os
 import sys
+import typing as T
 from collections.abc import Generator
 from hashlib import sha256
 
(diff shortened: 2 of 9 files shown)


test/test_pydot.py
```diff
@@ -105,11 +105,10 @@ def pydot(cls, filename: str, encoding: str) -> PydotRenderResult:
...
             raise RuntimeError("No data returned from pydot!")
-        jpe_data = bytearray()
-        for g in c:
-            jpe_data.extend(
-                g.create(prog=TEST_PROGRAM, format="jpe", encoding=encoding)
-            )
+        if len(c) > 1:
+            raise RuntimeError("Multiple graphs not supported!")
+        g = c[0]
+        jpe_data = g.create(prog=TEST_PROGRAM, format="jpe", encoding=encoding)
```
ferdnyc: Hmm. Seems we do have a test file with multiple graphs in it. Well, `dot` renders only the first one, so I guess I'll have the test code do the same.

ferdnyc:
Hrm. Now we have an API conflict here. These changes broke one of the tests:

* The `pydot.core.Node()` constructor accepts `int`, `float`, and `bool` as its `name=` argument, in addition to `str`. Whatever type the parameter is, it stores it in `obj_dict` as `str(name)`. (The latter was existing code.)
   Necessary because some of our own tests are calling pydot.Node(1) or pydot.Node(True) and expecting that to work.
* `pydot.core.Node.set_name()` now also accepts `int`, `float`, and `bool` in addition to `str`, and does NOT accept `None`. For parity with the constructor, it now also stores `str(node_name)` in the `obj_dict`.

...Because the output code has special support for `bool` values, outputting `True` as `true`. So, when `Node.set_name(True)` was called, that value used to be stored directly, and would then be output as `true`. Now that it's stored as `str(True)`, it's output as `True`.

Which means the API was already kind of broken, because, in `main`:
```python
>>> import pydot
>>> pydot.Node(True, color="red").to_string()
'True [color=red];'
>>> n = pydot.Node("a", color="red")
>>> n.set_name(True)
>>> n.to_string()
'true [color=red];'
```

since the Node constructor was doing `str(name)`, it makes no sense for `.set_name()` not to be doing the same thing. But apparently the code has support for non-`str` name values.

And, actually, I was wrong about "Whatever type the parameter is, it stores it in `obj_dict` as `str(name)`." In the constructor, that's only true for `isinstance(name, int)`, which covers `int` AND `bool` (because `isinstance(True, int)` is _TRUE_, crazily), but NOT `float`.

This is kind of a mess.

ferdnyc:
One relatively simple solution is to just have the `.set_name()` and `Node()` constructor store `str(name).lower()`, when `name` is a `bool`.

But I really don't like this, "store `int` as `str()`, but let `float` pass through untouched" thing the Node constructor is doing, and I really don't want to replicate bad logic as-is into `.set_name()` just for parity.

test/test_pydot.py
ferdnyc: Fixed.

ferdnyc: Not fixed. What I was apparently thinking is that `dot` does the _exact_ same thing: When you ask it to render multiple graphs to a binary format like `jpeg`, it just returns multiple images all glommed together.

That doesn't actually _work_, because what it ends up looking like to any software that supports the file format is: an image of the first graph, with N bytes of bad data appended to the end.

Anyway, I guess I'll have to go back to the `bytearray` method for the test, so that the outputs evaluate the same, and just cast it to `bytes` for typing purposes.

ferdnyc: Now really-fixed.

src/pydot/core.py
ferdnyc, line 372: My changes to the Node-name storage also lost coverage of this line, understandably. Even though, like the comment says a couple of lines down, "Based on type annotations, should never happen". But it was happening, in violation of what types we claimed our API accepts and returns.

ferdnyc: I think we should drop those lines from `quote_id_if_necessary()`, and change the classes so they actually store only `str` names, like they purport. Something like this, in both constructors and `.set_name()`:

```python
if isinstance(name, bool):
    self.obj_dict["name"] = str(name).lower()
else:
    self.obj_dict["name"] = str(name)
```

We may need to fix up our float handling to better support different values when stringified, though. Right now, this happens:
```python
>>> import pydot
>>> pydot.Node(-14.4).to_string()
'-14.4;'
>>> pydot.Node("-14.4").to_string()
'"-14.4";'
```

ferdnyc: One thing we could do, instead of using the messy regular expressions, is just try converting the value to `int()` or `float()` on **output**. For example:

```python
>>> int("-12")
-12
>>> float("-14.4")
-14.4
>>> int("12.2")
  File "<python-input-13>", line 1, in <module>
    int("12.2")
    ~~~^^^^^^^^
ValueError: invalid literal for int() with base 10: '12.2'
>>> float("7.5:sw")
  File "<python-input-14>", line 1, in <module>
    float("7.5:sw")
    ~~~~~^^^^^^^^^
ValueError: could not convert string to float: '7.5:sw'
>>> float('"7.5"')
  File "<python-input-16>", line 1, in <module>
    float('"7.5"')
    ~~~~~^^^^^^^^^
ValueError: could not convert string to float: '"7.5"'
```

So if `int(name)` succeeds, we output `name`. Otherwise, if `float(name)` succeeds, we output `name`. Only if both of those fail, do we bother with quoting logic. But screw the complicated regular expressions looking for numeric values.

(I say "output `name`" rather than "output `float(name)` or "output `str(float(name))`", because for example `float("-.5")` is `-0.5`, but `dot` supports `-.5` as a valid ID, and treats it _different_ from the also-valid `-0.5`. So, we don't want to _actually_ convert stringified `int`s or `float`s into those types, just see if we CAN.)

ferdnyc: Unfortunately `bool(any_str)` returns `True`. And `int("true")` is a `ValueError` (unlike `int(True)` which is `1`). So, not much help there, recognizing boolean values if they're stringified. But, `true` and `false` are valid strings we'd output unquoted anyway, so it shouldn't really matter.

ferdnyc: Also, if we _really_ want to be able to support all valid combinations of `id`, `id:port`, `id:compass`, and `id:port:compass`, with and without quoting, I think we may need to use a small pyparsing grammar to interpret the values. Regular expressions are never going to be able to recognize all of these things:

These are always true...
* `red` is `id="red"`
* `"red:5"` is `id="red:5"`
* `"red:5":"sw"` is `id="red:5"`, `compass="sw"`

_buuut_, assuming `red` has no port designated `ne`...
* `red:ne` is `id="red"`, `compass="ne"`
* `"red":"ne"` is `id="red"`, `compass="ne"`
* :face_with_spiral_eyes: `red:ne:sw` is `id="red"`, `compass="ne"` <sup>[1](#notes)</sup>
* :face_with_spiral_eyes: `red:ne:5` is `id="red"`, `compass="ne"` <sup>[1](#notes)</sup>
* :face_with_spiral_eyes: `red:"ne":5` is `id="red"`, `compass="ne"` <sup>[1](#notes)</sup>

Everything below here is technically only true if `red` has a port designated `5`, otherwise it breaks the endpoint (meaning if an edge uses the value, dot prints "Warning: node red, port 5 unrecognized" and loses the compass attachment also):
* :thinking: `red:5:sw` is `id="red"`, `port="5"`, `compass="sw"`
* :thinking: `"red":5:sw` is `id="red"`, `port="5"`, `compass="sw"`
* :thinking: `"red":5:"sw"` is `id="red"`, `port="5"`, `compass="sw"`
* :thinking: `"red":"5":"sw"` is `id="red"`, `port="5"`, `compass="sw"`

OTOH, if `red` has a port designated `sw`...
* :scream: `red:sw` is `id="red"`, `port="sw"`
* :scream: `red:sw:ne` is `id="red"`, `port="sw"`, `compass="ne"`
* :scream: `red:sw:5` is treated as `id="red"`, `port="sw"` but with a `dot` warning <sup>[2](#notes)</sup>
* :face_with_spiral_eyes: `red:ne:sw` is still `id="red"`, `compass="ne"` <sup>[1](#notes)</sup>

Even with a grammar... 

* Items marked with :scream: are basically impossible for us to interpret correctly, because we have no information about what ports a node does and doesn't have defined.
* Items marked with :thinking: can be interpreted "blindly", but we can't know that the graph rendering may be different depending on the validity/invalidity of the port value.
* Items marked :face_with_spiral_eyes: require us to examine the second position (first value after the colon) and check whether it's a valid compass direction, even in two-colon (three-value) forms. <sup>[1](#notes)</sup>

### Notes
1. `dot` will not complain if a two-colon endpoint to a node with no ports / no matching ports has a third value, valid or not, if the second value is a valid `compass` argument. It will simply treat it as the `compass` value, and ignore whatever follows.
2. The previous note is true _unless_ the second value IS defined as a port, in which case it's accepted as such and the third value becomes the compass direction. So, `red:sw:5` on a `red` with no ports is silently accepted as `red:sw`, but if `red` HAS a port `sw`, `dot` prints "Warning: node red, port sw, unrecognized compass point '5' - ignored"

ferdnyc: ^ Filed as #578.

ferdnyc:
Draft-ified mostly pending discussion of the API conflict I mentioned above.

PR #584, opened 2026-09-04
Test that all three DOT comment forms are discarded

ArsalLivoroi:
The `.dot` fixtures exercise `/* */` (21 files) and `//` (5 files)
incidentally, but no fixture contains a `#` line, and nothing asserts the
rule directly.

Adds one test covering all three forms.

### DIFF
--- a/test/test_parser.py
+++ b/test/test_parser.py
@@ -239,3 +239,23 @@ def test_bad_parse_bracket(capsys) -> None:
     assert res is None
     captured = capsys.readouterr()
     assert captured.out.strip() == expected
+
+
+def test_comments_are_discarded() -> None:
+    """Test that all three comment forms are ignored.
+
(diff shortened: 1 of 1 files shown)


test/test_parser.py
```diff
@@ -11,7 +11,7 @@
 import pyparsing as pp
 import pytest
 
-from pydot import dot_parser
+from pydot import dot_parser, graph_from_dot_data
```
lkk7: Thanks!

Could this test use `dot_parser.parse_dot_data()` like the other tests in this file? That will avoid the unnecessary import of `graph_from_dot_data`.

ArsalLivoroi: Done, thanks. I removed the extra import, so the diff now only covers the new test.

lkk7:
@ArsalLivoroi thank you!

